Issue #1, opened 2020-01-29
Restructure former role into collection


geerlingguy:
##### SUMMARY

Restructure this repository's structure from a role into a collection (it contains the historical commits / data from https://github.com/ansible/ansible-kubernetes-modules).

##### ISSUE TYPE

- Feature Idea

geerlingguy:
See: https://docs.ansible.com/ansible/latest/dev_guide/developing_collections.html#migrating-ansible-content-to-a-collection

geerlingguy:
And eventually, this collection will go into https://galaxy.ansible.com/community, so it will be `community.kubernetes`.

geerlingguy:
Using the following vars in `site.yaml` from the content_collector:

```
  vars:
    ansible_source_directory: ./ansible-2.9.4
    destination_directory: ~/Downloads/kubecollect
    collection_namespace: community
    collection_name: kubernetes
    collection_description: Kubernetes Collection for Ansible
    source_sub_directory: /clustering/k8s
```

Also ran into two issues with the content_collector:

  - https://github.com/ansible/content_collector/issues/20
  - https://github.com/ansible/content_collector/issues/21

geerlingguy:
I have most everything complete _so far_, and integration tests are now passing (though sanity is not)—working on that in #2.

geerlingguy:
Done for now; moving on to testing locally and documenting everything in #3.

PR #4, opened 2020-01-30
Issue #1: Restructure role into collection using content_collector.

geerlingguy:


.github/workflows/ansible-test.yml
```diff
@@ -0,0 +1,26 @@
+name: CI
+on:
+  - pull_request
+
+jobs:
+  integration:
```
gundalow: Any reason `sanity` isn't here? If it's broken/difficult I'd be interested in knowing, and maybe able to help fix.

geerlingguy: @gundalow - Ah, just noticed there are some sanity tests run on k8s, I missed those initially. I thought they were skipped for the k8s modules but obviously not :P

geerlingguy: Got sanity tests added. Though I'm getting a few failures that ansible/ansible doesn't seem to get. Will have to look into those but will do so separately.

PR #12, opened 2020-02-03
Add authors to galaxy.yml.

geerlingguy:
Since the git history of the ansible modules repository and ansible/ansible have diverged, it is hard to get an accurate picture of all those who have contributed and will continue contributing to the Kubernetes Collection.

I would like to get a somewhat accurate picture of those who have been and are still interested in maintaining Kubernetes content for OpenShift, so this PR is a stab at trying to get the authors field in galaxy.yml to be accurate.

Please make sure the URL listed by your username is accurate for what you'd like to display. Note that the authors information is not (to my knowledge) displayed in Galaxy's UI at this point.

### DIFF
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -1,4 +1,12 @@
-authors: []
+authors:
+  - chouseknecht (https://github.com/chouseknecht)
+  - geerlingguy (https://www.jeffgeerling.com/)
+  - maxamillion (https://github.com/maxamillion)
+  - jmontleon (https://github.com/jmontleon)
+  - fabianvf (https://github.com/fabianvf)
+  - willthames (https://github.com/willthames)
+  - mmazur (https://github.com/mmazur)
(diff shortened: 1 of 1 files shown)


geerlingguy:
According to @fabianvf I am probably missing some kubevirt folks, and looking around in the code, it looks like maybe I should add in:

  - @mmazur https://github.com/mmazur
  - @felixfontein https://github.com/felixfontein
  - @jamescassell https://github.com/jamescassell

Any others?

felixfontein:
@geerlingguy I didn't really work on (or used) the kubevirt modules; I probably only did some generic maintenance stuff which also affected some of them. So there's no need to add me :)

geerlingguy:
Added the others, and I think we're good to go on the authors field. If further changes need to be made, they can be made in subsequent PRs, and we'll refine the list as we go.

This should make the collection ready for a 0.9.0 release.

Issue #13, opened 2020-02-05
CI Builds started failing with "cannot import name '_distro'"


geerlingguy:
##### SUMMARY

CI builds (using ansible-base) started failing in the past day or two with "cannot import name '_distro'"

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

ansible-test

##### ANSIBLE VERSION
ansible-base devel

##### CONFIGURATION

N/A

##### OS / ENVIRONMENT

N/A

##### STEPS TO REPRODUCE

```yaml

```

##### EXPECTED RESULTS

Tests pass.

##### ACTUAL RESULTS

```
Run command: ansible-doc -t connection community.kubernetes.kubectl
ERROR: Command "ansible-doc -t connection community.kubernetes.kubectl" returned exit status 250.
>>> Standard Error
ERROR! Unexpected Exception, this is probably a bug: cannot import name '_distro'
```

```
Run command: importer.py
See documentation for help: https://docs.ansible.com/ansible/devel/dev_guide/testing/sanity/import.html
ERROR: Found 8 import issue(s) on python 3.6 which need to be resolved:
ERROR: plugins/module_utils/common.py:27:0: traceback: ImportError: cannot import name '_distro'
ERROR: plugins/module_utils/raw.py:29:0: traceback: ImportError: cannot import name '_distro'
ERROR: plugins/module_utils/scale.py:26:0: traceback: ImportError: cannot import name '_distro' (at plugins/module_utils/common.py:27:0)
ERROR: plugins/modules/k8s.py:279:0: traceback: ImportError: cannot import name '_distro' (at plugins/module_utils/raw.py:29:0)
ERROR: plugins/modules/k8s_auth.py:152:0: traceback: ImportError: cannot import name '_distro'
ERROR: plugins/modules/k8s_info.py:140:0: traceback: ImportError: cannot import name '_distro' (at plugins/module_utils/common.py:27:0)
ERROR: plugins/modules/k8s_scale.py:121:0: traceback: ImportError: cannot import name '_distro' (at plugins/module_utils/common.py:27:0)
ERROR: plugins/modules/k8s_service.py:172:0: traceback: ImportError: cannot import name '_distro' (at plugins/module_utils/common.py:27:0)
```

geerlingguy:
Looks like it's related to https://github.com/ansible-community/collection_migration/pull/277

geerlingguy:
It should start working again in ~24 hours.

[...]

PR #14, opened 2020-02-06
Add k8s_exec module to execute command through the API

TristanCacqueray:
This change adds a new module to execute command in a container
through the API:
https://docs.okd.io/latest/dev_guide/executing_remote_commands.html#protocol

Related: https://github.com/ansible/ansible/pull/55029
Closes: https://github.com/ansible-collections/kubernetes/issues/7


geerlingguy:
(Just noting that the test failures here are due to upstream issues in the ansible-base repository; the issues should hopefully be cleared up by end of week, and then one of the maintainers can re-run the GitHub Actions workflow to see if it passes.)

Also, I would like to have a test or two using `k8s_exec` added to the integration tests in this repo (just to make sure it's working and stays working with future changes).

fabianvf:
@geerlingguy how are we doing the tests? If it's the same way as ansible/ansible then it won't work because it's just spinning up an API server, no containers are actually being run. We should probably rewrite the tests to make use of `kind` or something like that so we can do full integration tests

geerlingguy:
@fabianvf - Ah, in that case yes. And I'd say https://github.com/ansible-collections/kubernetes/issues/10 is super-necessary in that case, to make our test coverage way nicer.

I know you had https://github.com/ansible/ansible/pull/61466 lined up in ansible/ansible... time to pull the trigger? (Though we'd still need the change for ansible-test to work—but I might be able to pummel that through).

fabianvf:
yeah, that could do it, we could also just move to using the official `kind` tool. Basically our test pipeline would then be

`install kind` -> `kind create cluster` -> `invoke tests`, I'd hazard a guess that a lot of what `ansible-test` gets you doesn't help us a ton since we have pretty simple host management and mostly just need to interact with a kubernetes API. In the long run I plan to make a `kind` molecule driver, but for now we could pretty easily just stand it up manually and either invoke molecule with the `delegated` driver or just run the test playbook by hand

geerlingguy:
@fabianvf - I like that approach (using official CNCF kind). Do you have time to work on that? Otherwise I may be able to work on it next week. Just don't want to duplicate efforts!

fabianvf:
@geerlingguy I'll be out on PTO next week so go for it! I'm happy to help out when I get back though. The new `cluster` scenario in the `operator-sdk` scaffolding uses this pattern (we currently use it with `kind` in our CI)

fabianvf:
@TristanCacqueray you could still do something like what I've done for the k8s_log module, where I added tests and verified locally but didn't import them, so they won't be run in CI: https://github.com/ansible-collections/kubernetes/pull/16/files#diff-31a24fedeb58002db40bc88400eec0feR1

TristanCacqueray:
@fabianvf alright, b4f2e88 does that. i haven't tested the task though, i'd like to wait for ci to validate the tasks.

geerlingguy:
@TristanCacqueray - If you get a chance, can you rebase, just to make sure tests pass? I'm still working on the new kind-cluster-based tests, but at least the tests should pass since #13 is resolved.

TristanCacqueray:
Is there an easy way to get to the failed task in those check results?

TristanCacqueray:
@geerlingguy @fabianvf the pod creation (needed to test k8s_exec) is failing because of `no nodes available to schedule pods`. Is this something I can fix or should I remove the integration test from the `full_test.yml` playbook?

geerlingguy:
@TristanCacqueray - Go ahead and remove from `full_test.yml`, but leave in the test tasks file. I'm going to try to get a cluster setup via molecule where these integration tests can go when they need a fully functional cluster running.

geerlingguy:
@TristanCacqueray - After #22 was merged, there's now a KinD cluster environment managed by Molecule—can you add your integration test there (add in [tasks](https://github.com/ansible-collections/kubernetes/tree/master/molecule/default/tasks), with a reference from [playbook.yml](https://github.com/ansible-collections/kubernetes/blob/master/molecule/default/playbook.yml) to it)?

TristanCacqueray:
@geerlingguy ok, ea0326e adds the integration test to molecule.

TristanCacqueray:
Test succeeded: https://github.com/ansible-collections/kubernetes/runs/444923875#step:8:1252

molecule/default/tasks/exec.yml
```diff
@@ -0,0 +1,50 @@
...
+        name: "{{ pod }}"
+        namespace: "{{ ns }}"
+      spec:
+        containers:
+          - name: sleeper
+            image: busybox
+            command: ["sleep", "infinity"]
+
+  block:
+    - name: Ensure that k8s-log namespace exists
```
geerlingguy: k8s-exec, or maybe just `{{ ns }}` so it's always accurate.

molecule/default/tasks/exec.yml
```diff
@@ -0,0 +1,50 @@
+---
+- vars:
+    ns: k8s-exec
+    pod: sleep-pod
+    pod_def:
```
geerlingguy: Elsewhere in tests, we spell out the full name, so maybe use `exec_namespace` (above) and `exec_pod_definition` for this variable?

molecule/default/tasks/exec.yml
```diff
@@ -0,0 +1,50 @@
...
+    - name: Show k8s_exec output
+      debug:
+        var: output
+
+    - name: Assert k8s_exec output is correct
+      assert:
+        that:
+          - "'nameserver' in output.stdout"
+
+    - name: Cleanup namespace
```
geerlingguy: Can you put this in the block's `always` (as the other task includes in this directory do), so it always cleans out the namespace even if there's a failure?

plugins/modules/k8s_exec.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+ANSIBLE_METADATA = {'metadata_version': '1.1',
+                    'status': ['preview'],
+                    'supported_by': 'community'}
+
+DOCUMENTATION = '''
+module: k8s_exec
+short_description: Execute command in Pod
+version_added: "2.10"
```
geerlingguy: Just wondering how this will work with the new world order, since this is a collection and it's technically not going into Ansible 2.10... I'm going to ask the Ansible team about this field.

geerlingguy: Let's make this "1.0.0" instead of "2.10" — seems like there _could_ be issues generating HTML docs for this, but worst case I think it will just hide the version_added field contents. And that's a bug we can work on getting fixed.

molecule/default/tasks/exec.yml
TristanCacqueray: Done

molecule/default/tasks/exec.yml
TristanCacqueray: Done

molecule/default/tasks/exec.yml
TristanCacqueray: Done

plugins/modules/k8s_exec.py
TristanCacqueray: Done

geerlingguy:
LGTM!

TristanCacqueray:
Finally, thanks! I haven't follow what is a collection, would you mind sharing the instructions how to use that module now? e.g. a playbook or ansible.cfg...

geerlingguy:
@TristanCacqueray - This project's README file has instructions for using the collection in your playbook—in Ansible 2.10, the k8s modules will be moving out of the ansible/ansible project and instead will come from this collection (though there will be a distribution of Ansible that includes them, copied over from this collection).

But basically, you can do the following to use the collection (and your shiny new module):

    ansible-galaxy collection install community.kubernetes

Then in your playbook, add the section:

```
- hosts: [hosts here]

  collections:
    - community.kubernetes

  tasks:
    - k8s_exec:
      ...
```

Alternatively you can drop the `collections` section and specify the 'FQCN' (Fully Qualified Collection Namespace) for the task reference, so instead of `k8s_exec`, it would be `community.kubernetes.k8s_exec`.

geerlingguy:
(Note that the above requires at least Ansible 2.8.)

TristanCacqueray:
@geerlingguy alright, that works, thanks again.

PR #16, opened 2020-02-06
Add k8s_log module

fabianvf:
##### SUMMARY
Adds the k8s_log module, which allows you to pull logs from any resource that supports the /log endpoint (mostly just Pods I think).

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME

k8s_log

##### ADDITIONAL INFORMATION

I've added a test (and verified it worked locally), but it requires a real kubernetes cluster, not just the api server, so I didn't include it from the existing test playbooks.

### DIFF
--- a/molecule/default/tasks/log.yml
+++ b/molecule/default/tasks/log.yml
@@ -1,0 +1,87 @@
+---
+- block:
+    - name: ensure that k8s-log namespace exists
+      k8s:
+        kind: Namespace
+        name: k8s-log
+
+    - name: create hello-world deployment
+      k8s:
+        wait: yes
+        definition:
+          apiVersion: apps/v1
+          kind: Deployment
+          metadata:
+            name: hello-world
+            namespace: k8s-log
+          spec:
+            selector:
+              matchLabels:
+                app: hello-world
+            template:
+              metadata:
+                labels:
+                  app: hello-world
+              spec:
+                containers:
+                - image: busybox
+                  name: hello-world
+                  command: ['sh']
+                  args: ['-c', 'while true ; do echo "hello world" && sleep 10 ; done']
+                restartPolicy: Always
+
+    - name: retrieve the log by providing the deployment
+      community.kubernetes.k8s_log:
+        api_version: apps/v1
+        kind: Deployment
+        namespace: k8s-log
+        name: hello-world
+      register: deployment_log
+
+    - name: verify that the log can be retrieved via the deployment
+      assert:
+        that:
+          - "'hello world' in deployment_log.log"
+          - item == 'hello world' or item == ''
+      with_items: '{{ deployment_log.log_lines }}'
+
+    - name: retrieve the log with a label selector
+      community.kubernetes.k8s_log:
+        namespace: k8s-log
+        label_selectors:
+          - 'app=hello-world'
+      register: label_selector_log
+
+    - name: verify that the log can be retrieved via the label
+      assert:
+        that:
+          - "'hello world' in label_selector_log.log"
+          - item == 'hello world' or item == ''
+      with_items: '{{ label_selector_log.log_lines }}'
+
+    - name: get the hello-world pod
+      k8s_info:
+        kind: Pod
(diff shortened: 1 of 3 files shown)


geerlingguy:
@fabianvf - If you get a chance, can you rebase, just to make sure tests pass? I'm still working on the new kind-cluster-based tests, but at least the tests should pass since #13 is resolved.

molecule/default/tasks/log.yml
```diff
@@ -0,0 +1,87 @@
...
+              spec:
+                containers:
+                - image: busybox
+                  name: hello-world
+                  command: ['sh']
+                  args: ['-c', 'while true ; do echo "hello world" && sleep 10 ; done']
+                restartPolicy: Always
+
+    - name: retrieve the log by providing the deployment
+      community.kubernetes.k8s_log:
```
geerlingguy: Technically not required since we are supplying `community.kubernetes` in the `collections` list on the playbook level; the other tests call the modules directly (without FQCN), so if we want to change to always calling the FQCN, maybe we do that in a separate issue/PR?

plugins/modules/k8s_log.py
```diff
@@ -0,0 +1,232 @@
...
+  returned: success
+log_lines:
+  type: list
+  description:
+  - The log of the object, split on newlines
+  returned: success
+'''
+
+
+from ansible.module_utils.k8s.common import KubernetesAnsibleModule, AUTH_ARG_SPEC
```
geerlingguy: Needs updating for the FQCN

molecule/default/tasks/log.yml
fabianvf: just a leftover from when I was testing it separately from everything else

plugins/modules/k8s_log.py
```diff
@@ -0,0 +1,233 @@
...
+    - Only one of I(name) or I(label_selector) may be provided
+  container:
+    description:
+    - Use to specify the container within a pod to grab the log from.
+    - If there is only one container, this will default to that container.
+    - If there is more than one container, this option is required.
+    required: no
+
+extends_documentation_fragment:
+  - k8s_auth_options
```
geerlingguy: Ah, this needs to be FCQN'ed too — see other modules for examples.

plugins/modules/k8s_log.py
```diff
@@ -0,0 +1,233 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2019, Fabian von Feilitzsch <@fabianvf>
```
djzager: Not sure if this is the license block you want.

fabianvf: I think it's in keeping with the way the other modules handle copyright, what would you suggest instead?

djzager: Had been expecting https://github.com/ansible-collections/kubernetes/blob/master/plugins/lookup/k8s.py#L1-L17

djzager: Take my comment back after having looked at all of the other modules.

fabianvf:
@geerlingguy think I fixed everything, ready for review again

plugins/modules/k8s_log.py
```diff
@@ -0,0 +1,238 @@
...
+ANSIBLE_METADATA = {'metadata_version': '1.1',
+                    'status': ['preview'],
+                    'supported_by': 'community'}
+
+DOCUMENTATION = '''
+module: k8s_log
+
+short_description: Fetch logs from Kubernetes resources
+
+version_added: "2.8"
```
geerlingguy: One more request, completely forgot about it in earlier reviews—can you set this to "1.0.0"? Apparently this won't break anything in ansible-doc, and would be more accurate since that will be the collection version this is added in.

fabianvf: done

geerlingguy:
LGTM, if tests pass we can merge it!

PR #20, opened 2020-02-11
Add daily scheduled CI workflow run at 6 a.m. UTC.

geerlingguy:


geerlingguy:
This way (especially right now, as there's a transition period from 2.9 to 2.10), we'll see at least by the day when a problem crops up in ansible-base that will affect this project's stability.

geerlingguy:
Documented the syntax here since GitHub Actions' docs didn't have much to say: [Running a Github Actions workflow on schedule and other events](https://www.jeffgeerling.com/blog/2020/running-github-actions-workflow-on-schedule-and-other-events)

PR #22, opened 2020-02-12
Use KinD cluster with Molecule for full integration tests

### LINKED ISSUE #10: Figure out what to do with OpenShift tests
##### SUMMARY

Currently, the tests run by `ansible-test` (e.g. `ansible-test integration --docker -v --color`) are running on OpenShift 3.9.0 (`openshift/origin:v3.9.0`), which has been out of support for some time.

The current release of OpenShift 3 is 3.11, though the Docker Hub image for that version is 2 years old (https://hub.docker.com/r/openshift/origin/tags), and also uses Kubernetes 1.11 as a base, which has not been supported upstream for some time either (see the [K8s version skew policy](https://kubernetes.io/docs/setup/release/version-skew-policy/)).

It would be good to use a supported image for the CI environment k8s cluster testing—for Operator SDK, the `bsycorp/kind` (see [tags](https://hub.docker.com/r/bsycorp/kind/tags)). There's also the official [SIG kind](https://kind.sigs.k8s.io), which I've successfully used on other ansible testing projects ([example](https://github.com/geerlingguy/ansible-for-kubernetes/blob/master/tests/cluster-aws-eks.sh#L32-L42)).

Ideally, we would have something from CRC or OKD that's equivalent to the single-container approach in `openshift/origin`, but I'm not sure if there's any timeline for something CI/local-friendly for OpenShift 4.x.

##### ISSUE TYPE

- Bug Report

##### COMPONENT NAME

ansible-test / CI tests

geerlingguy:
##### SUMMARY

Closes #10 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

CI

##### ADDITIONAL INFORMATION

N/A


geerlingguy:
@fabianvf / @marwatk - I'm trying to get the existing tests working on a KinD cluster in K8s 1.17, and it seems the only fail I'm currently getting is the new test that was added in https://github.com/ansible/ansible/pull/57418 to resolve https://github.com/ansible/ansible/issues/56640 — when running the tasks, the assert fails:

```
    TASK [assert that validate_certs caused a failure (and therefore was correctly translated to verify_ssl)] ***
    fatal: [localhost]: FAILED! => {
        "assertion": "output is failed",
        "changed": false,
        "evaluated_to": false,
        "msg": "Assertion failed"
    }
```

For now I think I might comment out those tasks and create a follow-up issue to look at it again, just so we can get the full test suite running in CI.

geerlingguy:
Currently breaking on GitHub with:

```
    TASK [Check that daemonset wait worked] ****************************************
    fatal: [localhost]: FAILED! => {
        "assertion": "updated_ds_pods.resources[0].spec.containers[0].image.endswith(\":3\")",
        "changed": false,
        "evaluated_to": false,
        "msg": "Assertion failed"
    }
```

Works locally, though :/

geerlingguy:
Now getting:

```
    TASK [Check that paused deployment wait worked] ********************************
257
    fatal: [localhost]: FAILED! => {"msg": "The conditional check 'condition.reason == \"DeploymentPaused\"' failed. The error was: An unhandled exception occurred while templating '{{ pause_deploy.result.status.conditions | json_query(\"[?type==`Progressing`]\") | first }}'. Error was a <class 'ansible.errors.AnsibleError'>, original message: template error while templating string: no filter named 'json_query'. String: {{ pause_deploy.result.status.conditions | json_query(\"[?type==`Progressing`]\") | first }}"}
```

I'll have to see where that filter's coming from, or find a different way to get at the data.

geerlingguy:
Two other ansible/ansible follow-ups as a result of this work:

  - [Add ability to install a collection ignoring dependencies](https://github.com/ansible/ansible/issues/67399)
  - [ansible-galaxy list is not listing certain valid roles](https://github.com/ansible/ansible/issues/67365)

geerlingguy:
Need to rebase and clean up the commits here... will do that soon.

PR #34, opened 2020-02-20
Fix failing CI tests by updating args in k8s_service

### LINKED ISSUE #33: Two sanity tests are failing in CI
##### SUMMARY

`ansible-test` sanity checks are currently failing in CI:

```
Running sanity test 'validate-modules' with Python 3.6
Run command: /usr/bin/python3.6 /root/ansible/test/lib/ansible_test/_data/sanity/validate-modules/validate-modules --format json --arg-spec plugins/modules/k8s.py plugins/modules/k8s_auth.py plugins/modules/k8s_exec.py plugins/modules/k8s_info.py plugins/modules/k8s_log.py plugins/modules/k8s_scale.py plugins/modules/k8s_service.py --collection ansible_collections/community/kubernetes
ERROR: Found 2 validate-modules issue(s) which need to be resolved:
ERROR: plugins/modules/k8s_service.py:0:0: mutually_exclusive-unknown: mutually_exclusive contains terms which are not part of argument_spec: apply
ERROR: plugins/modules/k8s_service.py:0:0: mutually_exclusive-unknown: mutually_exclusive contains terms which are not part of argument_spec: src
See documentation for help: https://docs.ansible.com/ansible/devel/dev_guide/testing/sanity/validate-modules.html
```

##### ISSUE TYPE

- Bug Report

##### COMPONENT NAME

CI / Documentation

geerlingguy:
Closes #33.﻿

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.retry
 .idea
 *.log
+__pycache__/
 
 # Galaxy artifacts.
 *.tar.gz
--- a/plugins/modules/k8s_service.py
+++ b/plugins/modules/k8s_service.py
@@ -214,7 +214,7 @@ SERVICE_ARG_SPEC = {
     'merge_type': {'type': 'list', 'elements': 'str', 'choices': ['json', 'merge', 'strategic-merge']},
     'selector': {'type': 'dict'},
     'src': {
-      'type': 'path',
+        'type': 'path',
     },
     'type': {
         'type': 'str',
--- a/plugins/modules/k8s_service.py
+++ b/plugins/modules/k8s_service.py
@@ -40,6 +40,15 @@ options:
     - definition
     - inline
     type: dict
+  src:
+    description:
+    - "Provide a path to a file containing a valid YAML definition of an object dated. Mutually
+      exclusive with I(resource_definition). NOTE: I(kind), I(api_version), I(resource_name), and I(namespace)
+      will be overwritten by corresponding values found in the configuration read in from the I(src) file."
+    - Reads from the local file system. To read from the Ansible controller's file system, use the file lookup
+      plugin or template lookup plugin, combined with the from_yaml filter, and pass the result to
+      I(resource_definition). See Examples below.
+    type: path
   state:
     description:
     - Determines if an object should be created, patched, or deleted. When set to C(present), an object will be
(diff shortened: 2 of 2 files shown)


geerlingguy:
@mmazur / @fabianvf - I added the relevant arg specs and docs to silence the warnings emitted by the sanity check... but I'm wondering—does the k8s_service module actually accept these parameters the same as the k8s module? The common `raw.py` module_util _does_ state the options are mutually exclusive, but I've only ever tested them with the `k8s` module, not `k8s_scale`...

willthames:
I don't really understand why `apply` and `src` need to be mutually exclusive - surely that's no different to doing `kubectl apply -f $src` ?

geerlingguy:
@willthames - See https://github.com/ansible-collections/kubernetes/blob/7b5c599f6d6df6c792c87c7bb8544abc024baee0/plugins/module_utils/raw.py#L101-L104 — `src` and `apply` are marked as mutually exclusive to `resource_definition` and `merge_type`, respectively (not to each other).

willthames:
Ah right, yes, that does indeed make more sense. 

In answer to your earlier question, while I'd say that k8s_service shouldn't need apply, I added apply functionality mainly because I got hit so hard with updating services with the k8s module - I don't know if k8s_service handles that better - there are no tests for k8s_service (which means it probably shouldn't have been accepted)

I don't know if it accepts src either - I thought it was a higher level abstraction than that

geerlingguy:
@willthames - Original PR with some discussion for background: https://github.com/ansible/ansible/pull/48872 — it seems that the module was mostly added as a convenience, but I agree, I'd never used it (nor noticed it was in Ansible) until I was moving it into this collection. I normally use the `k8s` module for most anything (besides `k8s_exec`, that's been extremely handy of late).

geerlingguy:
I'm going to go ahead and merge this, mostly to get tests back to green. If it turns out the k8s_service module needs further refinement to actually support the parameter structure it inherits... then we can open up follow-up bug reports and/or consider what to do with it later.

fabianvf:
@geerlingguy it probably should not accept all those `k8s` paramters, #36 to track

geerlingguy:
@fabianvf - Okay, in that case when we untangle #36 we can drop these parameters again.

PR #35, opened 2020-02-21
Wait conditions

willthames:
##### SUMMARY
Improve k8s Deployment and Daemonset wait conditions
    
Ensure that Deployments and Daemonsets properly await
all replicas to be available. Correctly handles the
subtle edge case when a service account no longer exists.
    
Note that this will dramatically slow Daemonset updates

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
k8s

### DIFF
--- a/molecule/default/tasks/apply.yml
+++ b/molecule/default/tasks/apply.yml
@@ -179,6 +179,162 @@
           - k8s_service_3.result.spec.ports | length == 1
           - k8s_service_3.result.spec.ports[0].port == 8081
 
+    - name: Insert new service port
+      k8s:
+        definition:
+          apiVersion: v1
+          kind: Service
+          metadata:
+            name: apply-svc
+            namespace: "{{ apply_namespace }}"
+          spec:
+            selector:
+              app: whatever
+            ports:
+              - name: mesh
+                port: 8080
+                targetPort: 8080
+              - name: http
+                port: 8081
+                targetPort: 8081
+        apply: yes
+      register: k8s_service_4
+
+    - name: Check ports are correct
+      assert:
+        that:
+          - k8s_service_4 is changed
+          - k8s_service_4.result.spec.ports | length == 2
+          - k8s_service_4.result.spec.ports[0].port == 8080
+          - k8s_service_4.result.spec.ports[1].port == 8081
+
+    - name: Remove new service port (check mode)
+      k8s:
+        definition:
+          apiVersion: v1
+          kind: Service
+          metadata:
+            name: apply-svc
+            namespace: "{{ apply_namespace }}"
+          spec:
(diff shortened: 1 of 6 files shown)


willthames:
Tests failing would be fixed by #34

molecule/default/tasks/apply.yml
```diff
@@ -179,6 +182,227 @@
...
+      k8s_scale:
+        api_version: apps/v1
+        kind: Deployment
+        name: scale-deploy
+        namespace: "{{ apply_namespace }}"
+        replicas: 0
+
+    - name: wait for pods to scale down
+      pause:
+        seconds: 30
```
fabianvf: can we use the `wait` field for this scale down?

plugins/module_utils/scale.py
```diff
@@ -29,11 +29,16 @@
...
-    from openshift import watch
     from openshift.dynamic.client import ResourceInstance
-    from openshift.helper.exceptions import KubernetesException
-except ImportError as exc:
-    class KubernetesException(Exception):
+    from openshift.dynamic.exceptions import NotFoundError
+except ImportError:
+    pass
+try:
+    from openshift import watch
```
fabianvf: I don't think we have a `watch` module in the openshift client anymore, we just make use of the `kubernetes` `watch` module instead. Only way to call a watch from the openshift client is the `DynamicClient.watch` method, though using `watch` from `kubernetes` should work fine.

plugins/module_utils/scale.py
```diff
@@ -189,15 +194,12 @@ def _create_stream(self, resource, namespace, wait_time):
...
-            w._api_client = self.client.client
-            if namespace:
-                stream = w.stream(resource.get, serialize=False, namespace=namespace, timeout_seconds=wait_time)
-            else:
-                stream = w.stream(resource.get, serialize=False, namespace=namespace, timeout_seconds=wait_time)
-        except KubernetesException:
-            pass
+        w = watch.Watch()
+        w._api_client = self.client.client
+        if namespace:
```
fabianvf: It looks like these two branches are identical, we can probably just remove this check.

molecule/default/tasks/apply.yml
willthames: I'm moving this to a separate PR for scale testing

plugins/module_utils/scale.py
willthames: I'm moving this to a separate PR for scale improvements

plugins/module_utils/scale.py
willthames: I'm moving this to a separate PR for scale testing

PR #39, opened 2020-02-25
🏗 Support BuildRequests

tumido:
##### SUMMARY
Add support for `BuildRequest` in `k8s`. 
Fixes: https://github.com/ansible/ansible/issues/52887
Replaces: https://github.com/ansible/ansible/pull/67697

##### ISSUE TYPE
- Bugfix/Feature Pull Request

##### COMPONENT NAME
Module: `k8s`

##### ADDITIONAL INFORMATION
Slight modification in `create_project_request` allows for creating any type of request if the subresource mapping is specified in `execute_module` when looking up the resource (this may be simplified if support for other subresources is needed).

Tasks like the one bellow are now supported:
```
- name: Test play
  hosts: localhost
  connection: local
  tasks:
  - name: Trigger a build
    k8s:
      namespace: <NAMESPACE>
      state: present
      definition:
        kind: BuildRequest
        apiVersion: build.openshift.io/v1
        metadata:
          name: <BUILD_CONFIG_NAME>
```

### DIFF
--- a/plugins/module_utils/raw.py
+++ b/plugins/module_utils/raw.py
@@ -178,6 +178,9 @@ class KubernetesRawModule(KubernetesAnsibleModule):
             if kind.endswith('List'):
                 resource = self.find_resource(kind, api_version, fail=False)
                 flattened_definitions.extend(self.flatten_list_kind(resource, definition))
+            elif kind == 'BuildRequest':
+                resource = self.find_resource('BuildConfig', api_version, fail=True).subresources['instantiate']
+                flattened_definitions.append((resource, definition))
             else:
                 resource = self.find_resource(kind, api_version, fail=True)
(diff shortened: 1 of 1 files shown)


tima:
@tumido Is BuildRequests an OCP4 thing? I can only find docs about this resource in OCP3, not OCP4.

tumido:
@tima yeah, that's correct. It seems like that's the case... I can't find anything about this resource on OCP4 as well.

tima:
We've move development of all OpenShift specific functions over to another collection/repo. A lot is changing with the OpenShift build pipeline that we are holding off on implementing something, but we are committed to doing so once things settle down. We are tracking this work with ansible-collections/community.okd#32.

PR #51, opened 2020-03-11
warn about disclosure when using certain options

bcoca:
##### SUMMARY
some options push data inline into cli, this is not good when using secrets and can lead to disclosure 

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request

##### COMPONENT NAME

kubectl

### DIFF
--- a/plugins/connection/kubectl.py
+++ b/plugins/connection/kubectl.py
@@ -65,6 +65,7 @@ DOCUMENTATION = """
       kubectl_extra_args:
         description:
           - Extra arguments to pass to the kubectl command line.
+          - Please be aware that this passes information directly on the command line and it could expose sensitive data.
         default: ''
         vars:
           - name: ansible_kubectl_extra_args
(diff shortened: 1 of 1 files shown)


plugins/connection/kubectl.py
```diff
@@ -65,6 +65,7 @@
       kubectl_extra_args:
         description:
           - Extra arguments to pass to the kubectl command line.
+          - Please be aware that this passes information directly on the command line and it could expose sensitive data.
```
shanemcd: Should you add the `We recommend using the file based authentication options instead.` here too?

shanemcd: Guess not? Looks like this was just merged.

fabianvf: my bad, though I think the password/api key are the only sensitive auth bits you can pass in plain text that you need to work around with file based auth, the other kubectl options are either just paths to files or not sensitive AFAIK

bcoca:
backport https://github.com/ansible/ansible/pull/68195

PR #61, opened 2020-03-22
Add helm, helm_info, and helm_repository modules

LucasBoisserie:
##### SUMMARY

Add helm, helm_info, and helm_repository modules to manage helm deployment inside kubernetes cluster

Fix: ansible/ansible#61546
Original PR on ansible repo: ansible/ansible#62450

##### ISSUE TYPE

- New Modules Pull Request

##### COMPONENT NAME
helm
helm_info
helm_repository

##### ADDITIONAL INFORMATION

Tests are enabled only for Helm 3
Helm 2 support has been dropped


README.md
```diff
@@ -27,6 +27,8 @@ Click on the name of a plugin or module to view that content's documentation:
     - [k8s_info](https://docs.ansible.com/ansible/latest/modules/k8s_info_module.html)
     - [k8s_scale](https://docs.ansible.com/ansible/latest/modules/k8s_scale_module.html)
     - [k8s_service](https://docs.ansible.com/ansible/latest/modules/k8s_service_module.html)
+    - [helm_cli](https://docs.ansible.com/ansible/latest/modules/helm_cli_module.html)
+    - [helm_cli_info](https://docs.ansible.com/ansible/latest/modules/helm_cli_info_module.html)
```
LucasBoisserie: Doc pages don't exist
is there another way to ref doc ?

geerlingguy: Right now, no. I would leave those links for now, and supposedly sometime in the coming weeks (months?) there should be some solution for documentation browsing for Collections, either on docs.ansible.com or on Galaxy... not sure where. See https://github.com/ansible-collections/kubernetes/issues/31

tests/integration/targets/helm/aliases
```diff
@@ -0,0 +1,2 @@
+cloud/openshift
```
geerlingguy: Note: For integration tests, we now have a molecule setup (https://github.com/ansible-collections/kubernetes/tree/master/molecule/default) which runs against a full Kind cluster, meaning you don't just have API access now, you can test anything against it (including Helm 2 / Tiller, I think).

Can you move all the integration tests to the molecule converge playbook (maybe group them under `tasks/helm` in a directory since there are so many)?

plugins/modules/helm_cli_info.py
```diff
@@ -0,0 +1,290 @@
...
+    description:
+      - Release name to manage
+    required: true
+    type: str
+    aliases: [ name ]
+  release_namespace:
+    description:
+      - Kubernetes namespace where the chart should be installed
+      - Can't be changed with helm 2
+    default: "default"
```
fabianvf: Just for consistency, the other modules don't set a default for namespace and instead set it as required. I think it makes sense to be explicit since most CLI utilities don't default to `default`, but rather a namespace determined by the current context, and since we're not matching that behavior defaulting to a specific namespace could trip up some users.

plugins/modules/helm_cli_info.py
```diff
@@ -0,0 +1,290 @@
...
+      - Ignored when is helm 3
+    type: str
+  tiller_namespace:
+    description:
+      - Namespace of Tiller
+      - Ignored when is helm 3
+    default: "kube-system"
+    type: str
+
+#Helm options
```
fabianvf: For the auth options, do you think you could make use of the shared [`AUTH_ARG_SPEC`](https://github.com/ansible-collections/kubernetes/blob/master/plugins/module_utils/common.py#L97) that the other modules use?

plugins/modules/helm_cli_info.py
```diff
@@ -0,0 +1,290 @@
...
+
+    if rc != 0:
+        module.fail_json(
+            msg="Failure when executing Helm command. Exited {0}.\nstdout: {1}\nstderr: {2}".format(rc, out, err),
+            command=get_command
+        )
+
+    # Helm 3 return "null" string when no values are set
+    if not is_helm_2 and out.rstrip("\n") == "null":
+        return yaml.safe_load('{}')
```
fabianvf: just a nit, but why not just return `{}` directly?

plugins/modules/helm_cli_info.py
LucasBoisserie: The shared spec offer too many options that helm doesn't manage
Helm has only 2 options to configure k8s auth (kube-context && kubeconfig)

plugins/modules/helm_cli_info.py
LucasBoisserie: Ok, i will change that

plugins/modules/helm_cli_info.py
LucasBoisserie: It was a temporary fix for helm3 during alpha but the issue is still here
Your suggestion is better, i will change that
Thanks

tatemz:
I'm excited about this PR and watching it for updates.

I'm testing this out locally and notice that we might need a way to add helm repos before updating the cache.

Thoughts?

plugins/modules/helm_cli_info.py
fabianvf: ah that makes sense. How does it behave if you are running the helm commands from inside a cluster? I assume it will make use of the incluster config, but perhaps that's just automatic?

plugins/modules/helm_cli.py
```diff
@@ -0,0 +1,593 @@
...
+                    'status': ['preview'],
+                    'supported_by': 'community'}
+
+DOCUMENTATION = '''
+---
+module: helm_cli
+short_description: Manages Kubernetes packages with the Helm package manager
+description:
+  - Install, upgrade, delete packages with the Helm package manager.
+version_added: "2.10"
```
Akasurde, line 19: This is not required.

plugins/modules/helm_cli.py
```diff
@@ -0,0 +1,593 @@
...
+  - "yaml (https://pypi.org/project/PyYAML/)"
+options:
+  binary_path:
+    description:
+      - The path of a helm binary to use.
+    required: false
+    type: path
+  chart_ref:
+    description:
+      - chart_reference on chart repository
```
Akasurde, line 34:
```suggestion
      - chart_reference on chart repository.
```

Full stop at the end of each line in the description.

plugins/modules/helm_cli.py
```diff
@@ -0,0 +1,593 @@
...
+      - Chart repository URL where to locate the requested chart
+    required: false
+    type: str
+  chart_repo_username:
+    description:
+      - Chart repository username where to locate the requested chart
+      - Required if I(chart_repo_password) is specified
+    required: false
+    type: str
+  chart_repo_password:
```
Akasurde, line 52: I would try to find another method to validate the chart repo username and password. Since this will expose sensitive information to the devices like `/proc/pid/cmdline`

plugins/modules/helm_cli.py
```diff
@@ -0,0 +1,593 @@
...
+    if is_helm_2:
+        inspect_command = command + " inspect chart " + chart_ref
+    else:
+        inspect_command = command + " show chart " + chart_ref
+
+    rc, out, err = module.run_command(inspect_command)
+    if rc != 0:
+        module.fail_json(
+            msg="Failure when executing Helm command. Exited {0}.\nstdout: {1}\nstderr: {2}".format(rc, out, err),
+            command=inspect_command
```
Akasurde, line 373: This will expose chart repo username and password to std output and/or log aggregators like splunk or logstash and can lead to CVE.

plugins/modules/helm_cli.py
```diff
@@ -0,0 +1,593 @@
...
+        deploy_command += " --force"
+
+    if disable_hook:
+        deploy_command += " --no-hooks"
+
+    if release_values != {}:
+        try:
+            import tempfile
+        except ImportError:
+            module.fail_json(msg="Could not import the tempfile python module. Please install `tempfile` package.")
```
Akasurde, line 401: Could you please use `missing_required_lib` from basic.py?

plugins/modules/helm_cli.py
```diff
@@ -0,0 +1,593 @@
...
+
+        ],
+        required_together=[
+            ['chart_repo_username', 'chart_repo_password']
+        ],
+        supports_check_mode=True,
+    )
+
+    if not HAS_YAML:
+        module.fail_json(msg="Could not import the yaml python module. Please install `yaml` package.")
```
Akasurde, line 484: Use `missing_required_lib` from basic.py.

plugins/modules/helm_cli.py
```diff
@@ -0,0 +1,593 @@
...
+    elif release_state == "present":
+
+        if chart_version is not None:
+            helm_cmd += " --version=" + chart_version
+
+        if chart_repo_url is not None:
+            helm_cmd += " --repo=" + chart_repo_url
+            if chart_repo_username is not None and chart_repo_password is not None:
+                helm_cmd += " --username=" + chart_repo_username
+                helm_cmd += " --password=" + chart_repo_password
```
Akasurde, line 553: Is there any other way to specify `password` rather than the current implementation?

plugins/modules/helm_cli_info.py
```diff
@@ -0,0 +1,289 @@
...
+                    'status': ['preview'],
+                    'supported_by': 'community'}
+
+DOCUMENTATION = '''
+---
+module: helm_cli_info
+short_description: Get informations from Helm package deployed inside the cluster
+description:
+  -  Get informations (values, states, ...) from Helm package deployed inside the cluster
+version_added: "2.10"
```
Akasurde, line 19:
```suggestion
```

plugins/modules/helm_cli_info.py
```diff
@@ -0,0 +1,289 @@
...
+            helm_cmd_common += " --host=" + tiller_namespace
+        else:
+            helm_cmd_common += " --tiller-namespace=" + tiller_namespace
+
+    if kube_context is not None:
+        helm_cmd_common += " --kube-context " + kube_context
+
+    if kubeconfig_path is not None:
+        helm_cmd_common += " --kubeconfig " + kubeconfig_path
+
```
Akasurde, line 279: Can we deprecate https://github.com/ansible-collections/community.general/blob/master/plugins/modules/cloud/misc/helm.py at the same time when we merge these modules?

plugins/modules/helm_cli.py
fabianvf: It looks like it may be possible to use the `helm registry login` subcommand to pipe the credentials in instead of adding them directly, though I'm not super familiar with how it works https://helm.sh/docs/topics/registries/#login

plugins/modules/helm_cli_info.py
LucasBoisserie: I just test inside a pod and it used the SA. Nothing is explained inside the helm documentation so I check helm's source code. 
It seems to load the incluster config with `k8s.io/cli-runtime` package from env vars.

plugins/modules/helm_cli_info.py
LucasBoisserie: PR created: https://github.com/ansible-collections/community.general/pull/115

plugins/modules/helm_cli_info.py
LucasBoisserie: Removed in last commit

plugins/modules/helm_cli.py
LucasBoisserie: Fixed in last commit

plugins/modules/helm_cli.py
LucasBoisserie: Fixed in last commit

plugins/modules/helm_cli.py
LucasBoisserie: Updated in last commit

plugins/modules/helm_cli.py
LucasBoisserie: Removed in last commit

plugins/modules/helm_cli.py
LucasBoisserie: I closed this conversation to keep only one with fabianvf

plugins/modules/helm_cli.py
LucasBoisserie: I closed this conversation to keep only one with fabianvf

plugins/modules/helm_cli.py
LucasBoisserie: `helm registry login` is available only for helm 3
`username` and `password` are used only for basic auth on the chart repository.
A alternative solution will be to make a `helm repo add` with the username and password before calling the module.
Helm will store these credentials inside its config files and will inject it automatically when the module will run commands.

Maybe the better solution is to remove these 2 arguments and put in the doc to run a `helm repo add` before ?

geerlingguy:
I was looking into what kind of support cycle there will be for Helm 2, because it would be nice to make a clean break (for some of the reasons @fabianvf pointed out with auth, mainly) and not have to support both 2 and 3...

Looking in the release notes for the last 2.x version, it states:

> 6 months after Helm 3’s public release, Helm 2 will stop accepting bug fixes. Only security issues will be accepted.
> 
> 12 months after Helm 3’s public release, support for Helm 2 will formally end. Download links for the Helm 2 client through Google Cloud Storage, the Docker image for Tiller stored in Google Container Registry, and the Google Cloud buckets for the stable and incubator chart repositories may no longer work at any point.

So, Helm 3 was released on November 13, 2019, meaning:

  - Helm 2 bug fixes until: May 13, 2020
  - Helm 2 security fixes until: November 13, 2020

Since this is a new module, and Helm 2 is already on the way out... what do others think about only supporting Helm 3 out of the gate?

LucasBoisserie:
@geerlingguy 

Helm 2 bug fixes has been prolonged to 13 August but I'm agree to drop helm 2 support.
When i start writting this module in october 2019, helm 3 was in beta, so i begin with helm 2
Now, i only used it with helm 3

For the auth part, i think the best solution is to outsource it and let users manage it.
There are many way to do it:
	* `helm registry` commands are only for OCI registry and are experimentals.
	* `helm repo add --username --password ` for helm repo.

We can mention them in the doc for more precision

xeor:
Agree that this should only be for `helm 3`. I don't think there are many people that will end up wanting to use this for `helm 2` since the same people would probably already be on `v3`, or days from it.
If they are using `v2`, they should follow https://helm.sh/docs/topics/v2_v3_migration/ to get to `v3` before using this.

LucasBoisserie:
CI seems to be broken
Sanity and integration failed with the same errors:
```
$ ansible-test sanity --docker -v --color --python 3.6
...
ERROR: Command "ansible-doc -t connection community.kubernetes.kubectl" returned exit status 1.
>>> Standard Error
Traceback (most recent call last):
  File "/root/ansible/test/lib/ansible_test/_data/injector/ansible-doc", line 79, in <module>
    main()
  File "/root/ansible/test/lib/ansible_test/_data/injector/ansible-doc", line 49, in main
    args += [find_executable(name)]
  File "/root/ansible/test/lib/ansible_test/_data/injector/ansible-doc", line 75, in find_executable
    raise Exception('Executable "%s" not found in path: %s' % (name, path))
Exception: Executable "ansible-doc" not found in path: /tmp/python-g6s_1c7y-ansible:/root/ansible/test/lib/ansible_test/_data/injector:/root/ansible/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
...
```

After investigation, the issue provide from ansible-devel, installed from this task:
```
- name: Install ansible base (devel branch)
  run: pip install https://github.com/ansible/ansible/archive/devel.tar.gz --disable-pip-version-check
```
Ansible 2.9 seems to have no issue
ref: https://github.com/ansible-collections/kubernetes/issues/63

geerlingguy:
@LucasBoisserie - Thanks for the update—it looks like there's a fix upstream already for ansible/ansible base, but the current GitHub outage is causing some delays in releasing it :(

Once that's through, someone can kick a re-test. I'm hoping we can get this merged before the end of the week!

geerlingguy:
@LucasBoisserie - For some reason now GitHub actions isn't even able to check out the code. I wonder if maybe one of your latest pushes didn't reach the repo correctly due to instability this morning (or if GitHub Actions is still experiencing weird outages)?

geerlingguy:
@LucasBoisserie - I just fixed the annoying GitHub Actions issue that is causing the new failures with your latest commit (see https://github.com/ansible-collections/kubernetes/pull/78) — can you rebase your branch and push it to kick off a new run using the new CI configuration?

LucasBoisserie:
Thanks @geerlingguy, the CI works well !

molecule/default/roles/helm/tasks/tests_chart.yml
```diff
@@ -0,0 +1,201 @@
...
+- name: Create helm namespace
+  k8s:
+    api_version: v1
+    kind: Namespace
+    name: "{{ helm_namespace }}"
+    wait: true
+
+- name: Check helm_cli_info empty
+  helm_cli_info:
+    binary_path: "{{ helm_binary}}"
```
geerlingguy, line 11: Missing space before closing brackets.

molecule/default/roles/helm/tasks/run_test.yml
```diff
@@ -0,0 +1,29 @@
+---
+- name: Ensure helm is not install
```
geerlingguy, line 2: s/install/installed

molecule/default/roles/helm/tasks/run_test.yml
```diff
@@ -0,0 +1,29 @@
...
+  file:
+    path: "{{ item }}"
+    state: absent
+  with_items:
+    - "/tmp/helm"
+
+- name: Check failed if helm is not install
```
geerlingguy, line 9: s/install/installed

molecule/default/roles/helm/tasks/test_helm_not_installed.yml
```diff
@@ -0,0 +1,15 @@
+---
+- name: Failed test when helm is not install
```
geerlingguy, line 2: s/install/installed

molecule/default/roles/helm/tasks/tests_chart.yml
```diff
@@ -0,0 +1,201 @@
...
+    name: test
+    namespace: "{{ helm_namespace }}"
+  register: empty_info
+
+- name: "Assert that no chart are installed with helm_cli_info"
```
geerlingguy, line 16: s/chart/charts

molecule/default/roles/helm/tasks/tests_chart.yml
geerlingguy: (Also on other instances of `helm_binary`)

molecule/default/roles/helm/tasks/tests_chart.yml
```diff
@@ -0,0 +1,201 @@
...
+
+- name: "Assert that {{ chart_test }} chart is upgraded with new var from {{ source }}"
+  assert:
+    that:
+      - install is changed
+      - install.status.status | lower == 'deployed'
+      - install.status.chart == "{{ chart_test }}-{{ chart_test_version }}"
+      - "install.status['values'].revisionHistoryLimit == 0"
+
+- name: Check idempotency after add vars
```
geerlingguy, line 84: s/add/adding

molecule/default/roles/helm/tasks/tests_chart.yml
```diff
@@ -0,0 +1,201 @@
...
+  helm_cli:
+    binary_path: "{{ helm_binary}}"
+    name: test
+    chart_ref: "{{ chart_source }}"
+    chart_version: "{{ chart_source_version | default(omit)}}"
```
geerlingguy, line 89: Missing space before closing brackets

geerlingguy: (also see other instances)

molecule/default/roles/helm/tasks/tests_chart.yml
```diff
@@ -0,0 +1,201 @@
...
+
+- name: "Assert that {{ chart_test }} chart is upgraded with new var from {{ source }}"
+  assert:
+    that:
+      - install is changed
+      - install.status.status | lower == 'deployed'
+      - install.status.chart == "{{ chart_test }}-{{ chart_test_version }}"
+      - install.status['values'] == {}
+
+- name: Check idempotency after remove vars
```
geerlingguy, line 119: s/remove/removing

molecule/default/roles/helm/tasks/tests_chart.yml
```diff
@@ -0,0 +1,201 @@
...
+  helm_cli:
+    binary_path: "{{ helm_binary}}"
+    name: test
+    chart_ref: "{{ chart_source }}"
+    chart_version: "{{ chart_source_version | default(omit)}}"
+    namespace: "{{ helm_namespace }}"
+  register: install
+
+- name: Assert idempotency after remove vars
```
geerlingguy, line 128: s/remove/removing

molecule/default/roles/helm/tasks/tests_chart.yml
```diff
@@ -0,0 +1,201 @@
...
+
+- name: "Remove {{ chart_test }} from {{ source }}"
+  helm_cli:
+    binary_path: "{{ helm_binary}}"
+    state: absent
+    name: test
+    namespace: "{{ helm_namespace }}"
+  register: install
+
+- name: "Assert that {{ chart_test }} chart is remove from {{ source }}"
```
geerlingguy, line 176: s/remove/removed

molecule/default/roles/helm/templates/tiller-clusterRoleBinding.yml
```diff
@@ -0,0 +1,13 @@
+---
```
geerlingguy, line 1: Are the tiller templates still required if we're not supporting Helm 2?

plugins/modules/helm_cli_info.py
```diff
@@ -0,0 +1,216 @@
...
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+ANSIBLE_METADATA = {'metadata_version': '1.1',
+                    'status': ['preview'],
+                    'supported_by': 'community'}
+
+DOCUMENTATION = '''
+---
+module: helm_cli_info
```
geerlingguy, line 15: Qq: Do we want to just call these `helm` and `helm_info` (instead of `helm_cli` and `helm_cli_info`)? Since this is entirely new content in the collection we don't have to worry about any old helm module...

tima: The one thing to consider against this would be forcing the use of the FQN of the module to avoid the helm module that is in core today and ACD in the future. Just a thought.

LucasBoisserie: It's ok for me
I already made a PR on ACD to put helm module in deprecated https://github.com/ansible-collections/community.general/pull/115
It will be removed in 4 releases

plugins/modules/helm_info.py
```diff
@@ -0,0 +1,216 @@
...
+  release_name:
+    description:
+      - Release name to manage.
+    required: true
+    type: str
+    aliases: [ name ]
+  release_namespace:
+    description:
+      - Kubernetes namespace where the chart should be installed.
+      - Can't be changed with helm 2.
```
geerlingguy, line 39: Can this be removed since we're not supporting Helm 2?

plugins/modules/helm.py
```diff
@@ -0,0 +1,494 @@
...
+    default: False
+    type: bool
+  wait_timeout:
+    description:
+      - Timeout when wait option is enabled (helm2 is a number of seconds, helm3 is a duration).
+    type: str
+'''
+
+EXAMPLES = '''
+- name: Create helm namespace HELM 3 doesn't create it automatically
```
geerlingguy, line 129: "Create namespace as Helm 3 doesn't create it automatically"

plugins/modules/helm.py
```diff
@@ -0,0 +1,494 @@
...
+  k8s:
+    api_version: v1
+    kind: Namespace
+    name: "monitoring"
+    wait: true
+
+# From repository
+- name: Add stable chart repo
+  shell: "helm repo add stable https://kubernetes-charts.storage.googleapis.com"
```
geerlingguy, line 138: Just wondering—should we add a module that allows managing chart repos more easily? Using shell is not idempotent unless the user does a lot of extra work. This could be a follow-up at some point.

plugins/modules/helm.py
```diff
@@ -0,0 +1,494 @@
...
+      replicas: 2
+
+- name: Deploy Grafana chart on 5.0.12 with values loaded from template
+  helm:
+    name: test
+    chart_ref: stable/grafana
+    chart_version: 5.0.12
+    values: "{{ lookup('template', 'somefile.yaml') | from_yaml }}"
+
+- name: Remove test release and waiting suppression ending
```
geerlingguy, line 155: "waiting suppression ending" — that seems unclear to me? Does it mean "wait until the release is removed?"

plugins/modules/helm.py
geerlingguy: (Ref @tatemz's comment from earlier in this PR thread)

> I'm testing this out locally and notice that we might need a way to add helm repos before updating the cache.

LucasBoisserie: I added a module to manage helm repository and update examples and tests

plugins/modules/helm.py
LucasBoisserie: This example is to illustrate the 'wait' option
Without the wait option helm only remove the objects contain in the release and not waiting that depending objects are removed (ex: deployment and pod)

README.md
```diff
@@ -27,6 +27,8 @@ Click on the name of a plugin or module to view that content's documentation:
     - [k8s_info](https://docs.ansible.com/ansible/latest/modules/k8s_info_module.html)
     - [k8s_scale](https://docs.ansible.com/ansible/latest/modules/k8s_scale_module.html)
     - [k8s_service](https://docs.ansible.com/ansible/latest/modules/k8s_service_module.html)
+    - [helm](https://docs.ansible.com/ansible/latest/modules/helm_module.html)
+    - [helm_info](https://docs.ansible.com/ansible/latest/modules/helm_info_module.html)
```
geerlingguy, line 31: Needs `helm_repository` added.

geerlingguy:
I think this PR's been through the ringer, but is quite ready for a merge. Barring any further objections, I think it would be best to merge it and release a new collection version (0.11.0) by the end of this week.

PR #84, opened 2020-05-04
K8s apply check mode fix

willthames:
##### SUMMARY
Adds test for applying a Service change in check_mode and ensures that test passes

Fixes ansible/ansible#66780

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
k8s/raw

##### ADDITIONAL INFORMATION
Needs backporting to ansible/ansible for 2.9

### DIFF
--- a/molecule/default/molecule.yml
+++ b/molecule/default/molecule.yml
@@ -21,6 +21,8 @@ provisioner:
         ansible_python_interpreter: '{{ ansible_playbook_python }}'
   env:
     ANSIBLE_FORCE_COLOR: 'true'
+  options:
+    vvv: True
 scenario:
   name: default
   test_sequence:
--- a/molecule/default/tasks/apply.yml
+++ b/molecule/default/tasks/apply.yml
@@ -154,6 +154,30 @@
         that:
           - k8s_service_2 is not changed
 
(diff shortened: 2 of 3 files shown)


willthames:
There is an intermittent test failure (there is a race condition) that is fixed by #87

molecule/default/molecule.yml
```diff
@@ -21,6 +21,8 @@ provisioner:
         ansible_python_interpreter: '{{ ansible_playbook_python }}'
   env:
     ANSIBLE_FORCE_COLOR: 'true'
+  options:
+    vvv: True
```
geerlingguy, line 25: I usually use default verbosity on test runs (when not specifically debugging an issue), but that's just me. @willthames and @fabianvf especially—do you prefer having more verbosity? I can go either way.

willthames: Definitely prefer more verbosity - lets me validate far more assumptions (which python is in use, which module is actually being exercised) and shows the full invocation, diff and results in the output.

fabianvf: I prefer greater verbosity as well, annoying to have to rerun a failed build by hand when the cause of an error is unclear

PR #99, opened 2020-05-14
Fix dry runs (--check) erroneously reporting changes

shaunsmiley-xevo:
##### SUMMARY

Fix --check mode reporting changes when there are none.

These issues are possibly related: 
Previously - 
- https://github.com/ansible/ansible/issues/66870
- https://github.com/ansible/ansible/issues/67653

Now - 
https://github.com/ansible-collections/community.kubernetes/issues/189
https://github.com/ansible-collections/community.kubernetes/issues/190

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

raw.py

##### ADDITIONAL INFORMATION

We know that we can run an ansible playbook using the k8s module over and over and see no changes (i.e. reports OK, not CHANGED).  However, we decided to add an additional step to validate which k8s tasks would change and what the changes were before executing with --check mode.  

We found that when we turned --check mode on from the cli, the k8s module reported changes every time, even when there would be no changes.  We confirmed this by running again without check mode seeing everything show OK (for 20+ different k8s manifests) then seeing everything show CHANGED with --check mode enabled. 

This fix also required a fix in the openshift lib.  That PR is here: https://github.com/openshift/openshift-restclient-python/pull/360

We are using both patches locally, and have tested internally that it is now working as expected.


geerlingguy:
@shaunsmiley-xevo - Would https://github.com/ansible-collections/community.kubernetes/pull/84 resolve this issue as well?

willthames:
@fabianvf given that this passes all test cases, we should probably ensure we have more test cases for check mode

mgogoi:
> @shaunsmiley-xevo - Would #84 resolve this issue as well?

We tested this and https://github.com/ansible-collections/community.kubernetes/pull/84 didn't fix the dry run issue.

mgogoi:
> Since this will require an updated version of the `openshift` client to work properly, would it be possible to fall back to the old behavior unless you have `openshift >= 0.11.1`?
> 
> edit: Also I believe dryRun was added in kubernetes 1.18, so we should probably fall back to the old behavior for kubernetes versions `< 1.18` as well

We are working on a fix to revert back to the old logic if the `openshift version < 0.11.1`.
The dry run feature in kubernetes has been there for a while now (since v1.13) - [https://v1-14.docs.kubernetes.io/docs/reference/using-api/api-concepts/#dry-run]. So, I feel its probably ok to pass `dry_run` to openshift,

fabianvf:
@mgogoi makes sense, thanks a ton!

willthames:
My thoughts on this are that this is usually a bug in our implementation of apply in the openshift library and all such bugs should be eliminated there (or here where it's not a bug in the openshift library). 

If there are outstanding test cases, then I'd be interested to see them (the test suite for apply is getting more and more extensive as I eliminate such bugs as I find them).

I'm already running `ansible-playbook -CDvvv kubernetes-playbook.yml` for a suitably large set of resources as a validation of this approach and finding very few false positive changes these days.

shaunsmiley-xevo:
@mgogoi's PR on our fork to revert behavior if openshift version is < 0.11.1 has been tested on our end and merged.  It is now included in this PR.

fabianvf:
@willthames I think if we can use the dry run API it makes sense to, it can only be more accurate than whatever we do client side

willthames:
Can we have some tests that exercise this? Not really sure how best to test the matrix of old and new openshift versions and old and new kubernetes clusters automatically but the tests would be a start, can always run the variations manually.

mgogoi:
@willthames - I am a bit unsure how to execute those tests. Could you please help me run those before I could add some new ones?

willthames:
@mgogoi sure, happy to help. Basically you need any kubernetes cluster (e.g. docker desktop, minikube, KinD) up and running. Install `molecule` however best suits your python setup (mine is in a virtualenv called ansible-kubernetes-testing). Then just run `molecule test` - it will pick up the kubernetes cluster from your current kubeconfig context

maxamillion (approved):

PR #100, opened 2020-05-16
k8s_scale fixes

willthames:
##### SUMMARY
Fix scale wait and add tests
    
Move wait logic out of raw and into common and use that
logic in scale
    
Fix a few broken wait condition cases highlighted by scaling
up and down
    
Move scale-related tests into dedicated test task file
##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
k8s_scale
##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,7 +76,7 @@ jobs:
           path: ansible_collections/community/kubernetes
 
       - name: Set up KinD cluster
-        uses: engineerd/setup-kind@v0.3.0
+        uses: engineerd/setup-kind@v0.4.0
 
       - name: Set up Python ${{ matrix.python_version }}
         uses: actions/setup-python@v1
--- a/molecule/default/converge.yml
+++ b/molecule/default/converge.yml
@@ -21,6 +21,7 @@
         that: (pod_list.resources | count) > 5
 
     - include_tasks: tasks/delete.yml
+    - include_tasks: tasks/scale.yml
     - include_tasks: tasks/apply.yml
     - include_tasks: tasks/waiter.yml
     - include_tasks: tasks/full.yml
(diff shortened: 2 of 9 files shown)


willthames:
Based on #35 - merge that one first

geerlingguy:
Latest test is giving:

```
Failed to apply object: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure",
"message":"Service "apply-svc" is invalid: spec.ports[1].nodePort: Forbidden: may not be used when `type` is 'ClusterIP'"
```

...which is weird, because it seems like `type` is set to `NodePort`...?

willthames:
@geerlingguy the bit where it was failing was in adding a port to an existing service, and also setting NodePort at the same time (I suspect I mis-merged two separate changes at some point) - just need to be consistently all NodePorts or all ClusterIPs. I've changed it to the latter.

geerlingguy:
Now this is failing with the same thing mentioned in #104 — 

```
    TASK [Check that deployment wait worked] ***************************************
    task path: /home/runner/work/community.kubernetes/community.kubernetes/ansible_collections/community/kubernetes/molecule/default/tasks/waiter.yml:247
fatal: [localhost]: FAILED! => {
    "assertion": "updated_deploy_pods.resources[0].spec.containers[0].image.endswith(\":2\")",
    "changed": false,
    "evaluated_to": false,
    "msg": "Assertion failed"
}
    Read vars_file 'vars/main.yml'
```

willthames:
Yeah, there's definitely a race condition, not 100% sure on the best course of action - Kubernetes is doing the right thing, the tests are doing the right thing, just perhaps too soon.

I might just have to improve the filters to ignore the unready pod where its container has terminated but the pod is still not in terminating state (that bit is weird)

geerlingguy:
Merging—it also seems to fix the instability with our CI tests, so I can stop getting nightly fail notifications from the scheduled CI run :)

PR #105, opened 2020-05-27
Use improved kubernetes diffs where possible

willthames:
##### SUMMARY
Use information from strategic merges to improve
the quality of diffs, particularly for list items
(order no longer important, and adding/removing an
item from a list just shows that item in the diff,
not the whole list)

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
common


willthames:
For this to be useful, https://github.com/openshift/openshift-restclient-python/pull/366 needs to be merged and released, but this will behave no differently to currently until that time

PR #107, opened 2020-05-28
Add action groups for playbooks with module_defaults

s-hertel:
##### SUMMARY
Keep backwards compatibility with playbooks using community.kubernetes content with module_defaults groups - for example

```
module_defaults:
  group/k8s:
    # set common options with default values
    option: value
tasks:
  # uses the module_default group value for the option
  - k8s_info:

  # should use the module_default group value for the option
  - community.kubernetes.k8s_info:
```

ansible/ansible#67291 adds support so the collection is no longer reliant on routing and the entries in core to keep things from breaking and now can define their own. This will also fix using `group/k8s:` with fully qualified module names.

##### ISSUE TYPE
- Bugfix Pull Request

### DIFF
--- a/meta/runtime.yml
+++ b/meta/runtime.yml
@@ -1,0 +1,12 @@
+action_groups:
+  k8s:
+  - k8s
+  - k8s_auth
+  - k8s_facts
+  - k8s_info
+  - k8s_scale
+  - k8s_service
+
+action_groups_redirection:
+  k8s:
+    redirect: k8s
--- a/meta/runtime.yml
+++ b/meta/runtime.yml
@@ -1,11 +1,12 @@
+---
 action_groups:
   k8s:
-  - k8s
-  - k8s_auth
-  - k8s_facts
-  - k8s_info
-  - k8s_scale
-  - k8s_service
+    - k8s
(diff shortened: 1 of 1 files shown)


geerlingguy:
@s-hertel - A few quick questions:

  1. Should all the k8s_* modules be under this action group? (There's also `k8s_log`, `k8s_exec` now).
  2. Should we have a 2nd action_group for the helm_ modules (there are three now, `helm`, `helm_info`, and `helm_repository`).
  3. Should the merging of this wait until https://github.com/ansible/ansible/pull/67291 is merged, or is that format pretty much set in stone/agreed upon? (I don't think it hurts anything to have it merged, I just don't want to merge this then have to re-format it again later.)
  4. Can you rebase the PR? There were some shaky tests that were recently fixed that should allow this to pass tests.

willthames:
Tests are still failing - looks like molecule's yamllint checks don't like the new meta/runtime.yml file.

My view is that I'm totally grateful for @s-hertel maintaining the existing status quo regarding k8s module defaults - whether we add missing k8s modules or helm modules is down to us as collection maintainers.

PR #118, opened 2020-06-03
Handle set object retrieved from lookup plugin

Akasurde:
##### SUMMARY

self.client.resources.api_groups is a dict_keys and is not
handled correctly by default callback plugin while JSON serialization.

This fix will typecast it to list so that it can be JSON serialized for
further processing.

Fixes: #111

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
molecule/default/tasks/full.yml
plugins/lookup/k8s.py

### DIFF
--- a/molecule/default/tasks/full.yml
+++ b/molecule/default/tasks/full.yml
@@ -185,6 +185,13 @@
           metadata:
             name: testing1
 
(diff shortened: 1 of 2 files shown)


fabianvf:
I probably should fix this in openshift/openshift-restclient-python too

PR #126, opened 2020-06-10
Issue #125: Add requires_ansible version constraints to runtime.yml.

### LINKED ISSUE #125: Add 'requires_ansible' requirements to collection
##### SUMMARY

It has been suggested we add `requires_ansible: '>=2.9.10,<2.11'` to `runtime.yml`.

Doing this more strongly indicates versions of Ansible core with which this collection should work. Honestly I would rather say `>=2.9` just to indicate that it works fine back to 2.9.0 (maybe with 2.8.0 but we don't test with that version so I'd rather not open a can of worms there).

##### ISSUE TYPE

- Feature Idea

##### COMPONENT NAME

`runtime.yml`

geerlingguy:
Closes #125.﻿

Leaving in draft state until we can choose whether to set the minimum requirement to `2.9.10` vs `2.9` (see related issue).

### DIFF
--- a/meta/runtime.yml
+++ b/meta/runtime.yml
@@ -1,4 +1,6 @@
 ---
+requires_ansible: '>=2.9,<2.11'
+
 action_groups:
   k8s:
     - k8s


geerlingguy:
From some discussions on IRC it seems that we're okay to go with 2.9 as the minimum version, since the content in this collection has been tested back that far.

PR #131, opened 2020-06-16
Issue #40: Add changelog and fragments and document new changelog process.

### LINKED ISSUE #40: Figure out the best way to handle CHANGELOG updates
##### SUMMARY

Ansible core uses 'changelog fragments' (see docs for [creating a changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment)) to make building a release-specific CHANGELOG file easier.

That might be overkill for a project with the reduced scope we have in this collection, but we still need a way to make sure all relevant changes (e.g. anything other than a grammar fix to the README or an adjustment to CI to fix failing tests) are encapsulated in the CHANGELOG file.

Right now, there are already two new module additions that are not documented in the CHANGELOG, but should be:

  - k8s_log: https://github.com/ansible-collections/kubernetes/pull/16
  - k8s_exec: https://github.com/ansible-collections/kubernetes/pull/14

Ideally, we would have some mechanism by which the CHANGELOG could be updated on a per-PR basis (and this would be one of the gates that would have to be passed before merging it)—but as with any other project where it's a requirement, one of the hardest parts is making sure users don't have to babysit their PRs while other ones come in and cause merge conflicts in the CHANGELOG file—there's no better way to discourage contributors than to make them do annoying tasks over and over again while their PR rots.

We could use a similar changelog generator that ansible/ansible uses, with sections including:

  - major_changes
  - minor_changes
  - deprecated_features
  - removed_features
  - bugfixes
  - known_issues

But we would need to integrate [`reno`](https://opendev.org/openstack/reno/) (or is it https://github.com/openstack/reno ?) into the release process (see [more reno docs](https://docs.openstack.org/reno/latest/user/usage.html#editing-a-release-note)).

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

CHANGELOG.md

##### ADDITIONAL INFORMATION

N/A

geerlingguy:
Closes #40.﻿


changelogs/.plugin-cache.yaml
```diff
@@ -0,0 +1,84 @@
+---
```
felixfontein, line 1: I'd `.gitignore` this file.

changelogs/changelog.yaml
```diff
@@ -0,0 +1,85 @@
...
+    fragments:
+    - 61-helm-new-modules.yaml
+    - 69-k8s_log-dont-parse-as-json.yaml
+    - 78-github-actions-workflow.yaml
+    - 81-rename-repository.yaml
+    - 85-exclude-unnecessary-files-when-building.yaml
+    modules:
+    - description: Manages Kubernetes packages with the Helm package manager
+      name: helm
+      namespace: .ansible.collections.ansible_collections.community.kubernetes.plugins.modules
```
felixfontein, line 61:
```suggestion
      namespace: ''
```
Same for other similar entries in this changelog.

changelogs/config.yaml
```diff
@@ -0,0 +1,30 @@
+---
+changelog_filename_template: CHANGELOG.rst
```
felixfontein, line 2:
```suggestion
changelog_filename_template: ../CHANGELOG.rst
```
This will *probably* place it one level up.

changelogs/.plugin-cache.yaml
geerlingguy: Fixed.

changelogs/config.yaml
geerlingguy: That seemed to work.

changelogs/changelog.yaml
geerlingguy: That seemed to work to patch over the issue; see comment in https://github.com/ansible-community/antsibull-changelog/issues/18.

geerlingguy:
Well that's annoying...

```
Command "ansible-doc -t connection community.kubernetes.kubectl" returned exit status 0.
>>> Standard Error
[WARNING]: Collection community.kubernetes does not support Ansible version
2.11.0.dev0
```

felixfontein:
Why not simply having `requires_ansible: '>=2.9`?

geerlingguy:
@fabianvf - Yeah... the thing is if I don't do that there's no changelog fragment and link back to the issue where the new module was added.

geerlingguy:
I just realized the issue ID doesn't make its way into the changelog at all... @fabianvf - should we manually drop a reference into each of the fragments too? I like being able to click through from GitHub's rendered page to an issue if I want to check out the issue/PR that introduced the change.

felixfontein:
@geerlingguy the recommendation in https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to is to include a link to the issue and/or PR in the fragment itself. That way you can also click on it when the community.kubernetes changelog is viewed as a subset of the Ansible changelog.

fabianvf:
@geerlingguy yeah I definitely like the PR reference in the changelog. On the operator-sdk we wrote a tool that combines fragments and finds the PR# for the change based on the PR referenced in the commit that merged the changelog fragment

geerlingguy:
@fabianvf - I think for now the idea is all the ACD/`pip install ansible` collections will build changelogs using `antisbull` so the fragments are compatible with the upstream build.

I've updated all the fragments to include PR/issue links, and regenerated the changelog. It seems everything's now ready to go!

Note: I didn't add any fragments for new issues/PRs merged since 0.11.0. That will (for now at least) be the responsibility of whomever tags the next release.

Issue #128, opened 2020-06-10
helm action doesn't push changes to local charts


iameli:
##### SUMMARY
If you make changes to a chart that's defined in a local directory, Helm won't deploy those changes. It will only attempt to kick off a deploy if the `values` change, not the chart definition itself.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
community.kubernetes.helm

##### ANSIBLE VERSION

```paste below
ansible 2.9.9
  config file = /Users/iameli/code/livepeer-infra/ansible/ansible.cfg
  configured module search path = ['/Users/iameli/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/Cellar/ansible/2.9.9/libexec/lib/python3.8/site-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 3.8.3 (default, Jun 10 2020, 15:04:14) [Clang 11.0.3 (clang-1103.0.32.29)]
```

##### CONFIGURATION

```paste below
HOST_KEY_CHECKING(/Users/iameli/code/livepeer-infra/ansible/ansible.cfg) = False
```

##### OS / ENVIRONMENT

macOS 10.15.3

##### STEPS TO REPRODUCE

1. Release a chart defined in a local directory
1. Make a change to any file within that chart (e.g. `example-chart/templates/example.yaml`)
1. Release the chart again

```yaml
- community.kubernetes.helm:
    chart_ref: "../charts/{{ livepeer_chart }}"
    release_name: "{{ livepeer_chart }}"
    kubeconfig_path: "../kubeconfigs/{{ inventory_hostname }}.secret"
    release_namespace: default
```

##### EXPECTED RESULTS
I'd expect the change to my local chart to be deployed to the cluster.

##### ACTUAL RESULTS

The module doesn't detect the change and nothing is released.

```paste below
TASK [helm : community.kubernetes.helm] *************************************************************************************
task path: /Users/iameli/code/livepeer-infra/ansible/roles/helm/tasks/main.yml:4
<localhost> ESTABLISH LOCAL CONNECTION FOR USER: iameli
<localhost> EXEC /bin/sh -c 'echo ~iameli && sleep 0'
[...]

PR #136, opened 2020-06-19
helm:  Support for single or multiple values files

### LINKED ISSUE #93: Helm support for single or multiple values files
##### SUMMARY

Could be interesting that the helm module will support single or multiple values files. 
Helm command itself support it adding "-f values.yaml -f values2.yaml"  
##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

* helm

##### ADDITIONAL INFORMATION

```yaml

```

ggee:
##### SUMMARY

Helm provides an option to give a list of yaml files that will be used to supply values in the order specified.  In this module, these value files will specified before the values: option generated file giving the values: in the task the final override.

Include "Fixes #93 "

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/helm.py

##### ADDITIONAL INFORMATION

```
    - name: Install nginx
      community.kubernetes.helm:
        chart_ref: helm-charts
        name: nginx
        namespace: testing
        values_files:
          - global_values.yaml
          - site_values.yaml
          - custom_values.yml

```

### DIFF
--- a/plugins/modules/helm.py
+++ b/plugins/modules/helm.py
@@ -81,6 +81,9 @@ options:
   values_files:
     description:
         - Value files to pass to chart.
+        - Paths will be read from the target host's filesystem, not the host running ansible.
+        - values_files option is evaluated before values option if both are used.
+        - Paths are evaluated in the order the paths are specified.
     required: false
     default: []
     type: list
--- a/plugins/modules/helm.py
+++ b/plugins/modules/helm.py
@@ -75,13 +75,14 @@ options:
     description:
         - Value to pass to chart.
     required: false
+    default: {}
     aliases: [ values ]
     type: dict
   values_files:
     description:
         - Value files to pass to chart.
     required: false
-    default: {}
+    default: []
     type: list
     elements: str
   update_repo_cache:
@@ -357,7 +358,7 @@ def deploy(command, release_name, release_values, chart_name, wait, wait_timeout
 
     if values_files:
         for value_file in values_files:
-            deploy_command += " -f=" + value_file
+            deploy_command += " --values=" + value_file
 
     if release_values != {}:
         fd, path = tempfile.mkstemp(suffix='.yml')
@@ -400,7 +401,7 @@ def main():
             release_namespace=dict(type='str', required=True, aliases=['namespace']),
             release_state=dict(default='present', choices=['present', 'absent'], aliases=['state']),
             release_values=dict(type='dict', default={}, aliases=['values']),
-            values_files=dict(type='list'),
+            values_files=dict(type='list', default=[], elements='str'),
             update_repo_cache=dict(type='bool', default=False),
 
             # Helm options
--- a/plugins/modules/helm.py
+++ b/plugins/modules/helm.py
@@ -78,6 +78,13 @@ options:
     default: {}
     aliases: [ values ]
     type: dict
+  values_files:
+    description:
+        - Value files to pass to chart.
+    required: false
+    default: {}
+    type: list
+    elemants: str
   update_repo_cache:
     description:
       - Run C(helm repo update) before the operation. Can be run as part of the package installation or as a separate step.
@@ -320,7 +327,7 @@ def fetch_chart_info(command, chart_ref):
     return yaml.safe_load(out)
 
 
-def deploy(command, release_name, release_values, chart_name, wait, wait_timeout, disable_hook, force):
+def deploy(command, release_name, release_values, chart_name, wait, wait_timeout, disable_hook, force, values_files):
     """
     Install/upgrade/rollback release chart
     """
@@ -339,6 +346,10 @@ def deploy(command, release_name, release_values, chart_name, wait, wait_timeout
 
     if disable_hook:
         deploy_command += " --no-hooks"
+
+    if values_files:
+      for value_file in values_files:
+        deploy_command += " -f=" + value_file
 
     if release_values != {}:
         try:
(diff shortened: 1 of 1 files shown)


plugins/modules/helm.py
```diff
@@ -75,9 +75,15 @@
     description:
         - Value to pass to chart.
     required: false
-    default: {}
```
Akasurde, line 78: Any reason why this is removed?

plugins/modules/helm.py
```diff
@@ -349,6 +355,10 @@ def deploy(command, release_name, release_values, chart_name, wait, wait_timeout
     if disable_hook:
         deploy_command += " --no-hooks"
 
+    if values_files:
+        for value_file in values_files:
+            deploy_command += " -f=" + value_file
```
Akasurde, line 360:
```suggestion
            deploy_command += " --values=" + value_file
```

plugins/modules/helm.py
```diff
@@ -75,9 +75,15 @@
...
         - Value to pass to chart.
     required: false
-    default: {}
     aliases: [ values ]
     type: dict
+  values_files:
+    description:
+        - Value files to pass to chart.
+    required: false
+    default: {}
```
Akasurde, line 84:
```suggestion
    default: []
```

plugins/modules/helm.py
```diff
@@ -390,6 +400,7 @@ def main():
             release_namespace=dict(type='str', required=True, aliases=['namespace']),
             release_state=dict(default='present', choices=['present', 'absent'], aliases=['state']),
             release_values=dict(type='dict', default={}, aliases=['values']),
+            values_files=dict(type='list'),
```
Akasurde, line 403:
```suggestion
            values_files=dict(type='list', default=[], elements='str'),
```

`elements='str'` make sure that we have only string values for `values_files`.

Akasurde: Also, this will solve CI failures - 

```
ERROR: Found 2 validate-modules issue(s) which need to be resolved:
ERROR: plugins/modules/helm.py:0:0: doc-elements-mismatch: Argument 'values_files' in argument_spec does not specify elements,but elements is documented as being str
ERROR: plugins/modules/helm.py:0:0: parameter-list-no-elements: Argument 'values_files' in argument_spec defines type as list but elements is not defined
```

plugins/modules/helm.py
ggee: no, mistake.  fixing.

Akasurde: Cool Thanks.

plugins/modules/helm.py
ggee: fixed

plugins/modules/helm.py
ggee: fixed

plugins/modules/helm.py
ggee: fixed

plugins/modules/helm.py
```diff
@@ -78,6 +78,13 @@
     default: {}
     aliases: [ values ]
     type: dict
+  values_files:
+    description:
+        - Value files to pass to chart.
```
fabianvf, line 83: Should we note that these paths will be read from the target host's filesystem, rather than the one that is running Ansible?

fabianvf: Also maybe note that order matters, and clarify what the precedence will be if I supply both `values` and `values_files`

ggee: Where would you like me to add this information?

fabianvf: Just in the parameter descriptions here, I think that's usually where this sort of info goes

fabianvf:
Awesome, changes all look good to me. @ggee would you mind adding some tests for this flag to the helm tests (https://github.com/ansible-collections/community.kubernetes/tree/master/molecule/default/roles/helm/tasks)

plugins/modules/helm.py
ggee: added

ggee:
> Awesome, changes all look good to me. @ggee would you mind adding some tests for this flag to the helm tests (https://github.com/ansible-collections/community.kubernetes/tree/master/molecule/default/roles/helm/tasks)

OK.  I haven't used molecule before and that directory doesn't seem to have a good example for local file access tests.  Any pointers?

fabianvf:
@ggee Yeah, molecule basically just runs the content as is, so if you want to access a file/template from the `helm` role, you can just put it into the `roles/helm/files` or `roles/helm/templates` directory (you probably need to create it), and you can access it like normal

plugins/modules/helm.py
```diff
@@ -471,13 +487,13 @@ def main():
 
         if release_status is None:  # Not installed
             helm_cmd = deploy(helm_cmd, release_name, release_values, chart_ref, wait, wait_timeout,
-                              disable_hook, False, atomic=atomic)
+                              disable_hook, False, values_files, atomic=atomic)
             changed = True
 
         elif force or release_values != release_status['values'] \
```
LucasBoisserie, line 493: You need to update the release_values variable otherwise idempotency will be broke
We looking to helm-diff in #128 to have a better management of idempotency

tima:
@ggee: What is the status of this PR? Looks like you need to incorporate suggestion from @fabianvf @LucasBoisserie to get this merged.

geerlingguy:
(Just noting I'm working on a rebase in https://github.com/ansible-collections/community.kubernetes/pull/236).

tima:
Closing this PR in favor of the rebase in #236 @geerlingguy has done.

PR #139, opened 2020-06-24
k8s: Fix a bug when manifest file ends with '---'

Akasurde:
##### SUMMARY

Any kubernetes manifest file ending with '---' will
generate an error when applied using 'k8s' module.

Although this may not be 'legal' YAML, this is quite frequent,
specially on helm's generated manifest files.

migrated from https://github.com/ansible/ansible/pull/59160

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
changelogs/fragments/59160_handle_none_def.yml
plugins/module_utils/raw.py

### DIFF
--- a/changelogs/fragments/59160_handle_none_def.yml
+++ b/changelogs/fragments/59160_handle_none_def.yml
@@ -1,0 +1,2 @@
+bugfixes:
+- raw - handle condition when definition is none (https://github.com/ansible/ansible/pull/59160).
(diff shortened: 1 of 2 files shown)


willthames:
It doesn't have to end in `---` for this to be a problem either - it can happen in the middle (again, another artifact of helm)

willthames:
Thanks @Akasurde

Issue #140, opened 2020-06-25
Helm module should use K8S_AUTH_KUBECONFIG when kubeconfig_path isn't specified


joao-fidalgo:
##### SUMMARY
Helm module should use K8S_AUTH_KUBECONFIG when kubeconfig_path isn't specified

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
helm

##### ADDITIONAL INFORMATION
This change would normalize behaviour between modules.

```yaml
- hosts: localhost
  gather_facts: false
  connection: local

  collections:
    - community.kubernetes

  environment:
    K8S_AUTH_KUBECONFIG: "/tmp/kubeconfig.yml"

  tasks:
    - name: install rancher using helm
       helm:
        chart_ref: rancher-stable/rancher
        name: rancher
        namespace: cattle-system
```

Akasurde:
Make sense to me. Working on it.

Akasurde:
@joao-fidalgo Could you please check if PR #141 works for you? Thanks

joao-fidalgo:
@Akasurde LGTM

PR #141, opened 2020-06-25
helm: Add support for K8S_AUTH_CONTEXT, K8S_AUTH_KUBECONFIG env

Akasurde:
##### SUMMARY

helm and helm_info modules can now use K8S_AUTH_KUBECONFIG and K8S_AUTH_KUBECONFIG
from environment variable if not specified in task.

Fixes: #140

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
changelogs/fragments/140_kubeconfig_env.yaml
plugins/modules/helm.py
plugins/modules/helm_info.py

### DIFF
--- a/plugins/modules/helm.py
+++ b/plugins/modules/helm.py
@@ -98,10 +98,12 @@ options:
   kube_context:
     description:
       - Helm option to specify which kubeconfig context to use.
+      - If the value is not specified in the task, the value of environment variable C(K8S_AUTH_CONTEXT) will be used instead.
     type: str
   kubeconfig_path:
     description:
       - Helm option to specify kubeconfig path to use.
+      - If the value is not specified in the task, the value of environment variable C(K8S_AUTH_KUBECONFIG) will be used instead.
     type: path
     aliases: [ kubeconfig ]
   purge:
@@ -245,7 +247,7 @@ except ImportError:
     IMP_YAML_ERR = traceback.format_exc()
     IMP_YAML = False
 
-from ansible.module_utils.basic import AnsibleModule, missing_required_lib
+from ansible.module_utils.basic import AnsibleModule, missing_required_lib, env_fallback
 
 module = None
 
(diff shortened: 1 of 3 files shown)


plugins/modules/helm.py
```diff
@@ -98,10 +98,12 @@
   kube_context:
     description:
       - Helm option to specify which kubeconfig context to use.
+      - If the value is not specified in the task, the value of environment variable C(K8S_AUTH_CONTEXT) will be used instead.
```
geerlingguy, line 101: I'm jusst now realizing... it seems these two values are inconsistent with the same values for the other k8s modules, e.g. https://github.com/ansible-collections/community.kubernetes/blob/98b72a4b2dacf447eab96dd6cbf09aa65825554d/plugins/doc_fragments/k8s_auth_options.py#L24-L34

Maybe we should file a follow-up issue to deprecate these parameters and use consistent parameters across all the modules?

plugins/modules/helm_info.py
```diff
@@ -177,8 +179,8 @@ def main():
             release_namespace=dict(type='str', required=True, aliases=['namespace']),
 
             # Helm options
-            kube_context=dict(type='str'),
-            kubeconfig_path=dict(type='path', aliases=['kubeconfig']),
+            kube_context=dict(type='str', fallback=(env_fallback, ['K8S_AUTH_CONTEXT'])),
+            kubeconfig_path=dict(type='path', aliases=['kubeconfig'], fallback=(env_fallback, ['K8S_AUTH_KUBECONFIG'])),
```
fabianvf, line 183: wow I wish I knew that existed

plugins/modules/helm.py
fabianvf: +1

Akasurde: I will work on follow-up PR.

PR #154, opened 2020-07-07
helm_plugin and helm_plugin_info: new module

Akasurde:
##### SUMMARY

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
molecule/default/roles/helm/tasks/run_test.yml
molecule/default/roles/helm/tasks/tests_helm_plugin.yml
plugins/modules/helm_plugin.py

### DIFF
--- a/plugins/modules/helm_plugin_info.py
+++ b/plugins/modules/helm_plugin_info.py
@@ -1,0 +1,167 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2020, Ansible Project
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = r'''
+---
+module: helm_plugin_info
+short_description: Gather information about Helm plugins
+version_added: "0.11.0"
+author:
+  - Abhijeet Kasurde (@Akasurde)
+requirements:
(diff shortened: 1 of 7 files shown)


plugins/modules/helm_plugin_info.py
```diff
@@ -0,0 +1,175 @@
...
+    required: false
+    type: path
+  release_namespace:
+    description:
+      - Kubernetes namespace where the helm plugins are installed.
+    required: true
+    type: str
+    aliases: [ namespace ]
+
+#Helm options
```
fabianvf, line 34: Not a blocker at all for this PR but I feel like we should break these common options into doc fragments etc to cut down on the repetition

plugins/modules/helm_plugin_info.py
```diff
@@ -0,0 +1,175 @@
...
+                continue
+            name, version, description = line.split('\t', 3)
+            if plugin_name is not None and plugin_name == name:
+                plugin_list.append({
+                    'name': name.strip(),
+                    'version': version.strip(),
+                    'description': description.strip(),
+                })
+                break
+            plugin_list.append({
```
fabianvf, line 159: Should we not do this when `plugin_name` has been set? I think currently the module will return info on all the plugins up to the one specified when `plugin_name` is set

plugins/modules/helm_plugin_info.py
Akasurde: I will make a separate PR for this. Thanks for the suggestions.

PR #155, opened 2020-07-08
Ensure check mode results are as expected

willthames:
##### SUMMARY
check_mode with wait updates the result based on the existing resource
which won't change with check_mode, so returns incorrect results.
Only run the wait checks in non check_mode (this also avoids the issue
of waiting for resource creation that will never happen in check mode)

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
raw

### DIFF
--- a/molecule/default/tasks/apply.yml
+++ b/molecule/default/tasks/apply.yml
@@ -324,6 +324,45 @@
             cpu: 100m
             memory: 100Mi
 
+    - name: Update the earlier deployment in check mode
+      k8s:
+        definition:
+          apiVersion: apps/v1
+          kind: Deployment
+          metadata:
+            name: apply-deploy
+            namespace: "{{ apply_namespace }}"
+          spec:
+            replicas: 1
+            selector:
+              matchLabels:
+                app: "{{ k8s_pod_name }}"
+            template: "{{ k8s_pod_template }}"
+        wait: yes
+        apply: yes
+      check_mode: yes
+      vars:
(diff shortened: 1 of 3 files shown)


molecule/default/tasks/apply.yml
```diff
@@ -324,6 +324,45 @@
...
+            namespace: "{{ apply_namespace }}"
+          spec:
+            replicas: 1
+            selector:
+              matchLabels:
+                app: "{{ k8s_pod_name }}"
+            template: "{{ k8s_pod_template }}"
+        wait: yes
+        apply: yes
+      check_mode: yes
```
geerlingguy, line 343: Might just be completely missing the point (though we've used this pattern a lot in the past), but since we're not running this playbook in check mode, this would just test that the change applies normally, right?

willthames: `check_mode: yes` means "when in normal mode, run this task in check mode". `check_mode: no` means "when in check mode, don't run this task"

willthames: https://docs.ansible.com/ansible/latest/user_guide/playbooks_checkmode.html#enabling-or-disabling-check-mode-for-tasks

geerlingguy: GAH you're right. For some reason my brain is always completely the opposite on that point.

I liked `always_run` way better, because I can never determine what happens by intuition with `check_mode`.

Issue #151, opened 2020-07-03
K8s_info module: error when using variable as app label selector


ElioLopez-md:
##### SUMMARY
The k8s_info module fails when using a variable as label selector.

##### ISSUE TYPE
- Bug Report

##### ANSIBLE VERSION

```paste below
ansible 2.9.6
  config file = None
  configured module search path = ['/Users/elopez/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/Cellar/ansible/2.9.6_1/libexec/lib/python3.8/site-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 3.8.3 (default, May 27 2020, 20:54:22) [Clang 11.0.3 (clang-1103.0.32.59)]
```

##### CONFIGURATION

```paste below
ACTION_WARNINGS(default) = True
AGNOSTIC_BECOME_PROMPT(default) = True
ALLOW_WORLD_READABLE_TMPFILES(default) = False
ANSIBLE_CONNECTION_PATH(default) = None
ANSIBLE_COW_PATH(default) = None
ANSIBLE_COW_SELECTION(default) = default
ANSIBLE_COW_WHITELIST(default) = ['bud-frogs', 'bunny', 'cheese', 'daemon', 'default', 'dragon', 'elephant-in-snake', 'elephant', 'eyes', 'hellokitty', 'kitty', 'luke-koala', 'meow', 'milk', 'moofasa', 'moose', 'ren', 'sheep', 'small', 'stegosaurus', 'stimpy', 'supermilker', 'three-eyes', 'turkey', 'turtle', 'tux', 'udder', 'vader-koala', 'vader', 'www']
ANSIBLE_FORCE_COLOR(default) = False
ANSIBLE_NOCOLOR(default) = False
ANSIBLE_NOCOWS(default) = False
ANSIBLE_PIPELINING(default) = False
ANSIBLE_SSH_ARGS(default) = -C -o ControlMaster=auto -o ControlPersist=60s
ANSIBLE_SSH_CONTROL_PATH(default) = None
ANSIBLE_SSH_CONTROL_PATH_DIR(default) = ~/.ansible/cp
ANSIBLE_SSH_EXECUTABLE(default) = ssh
ANSIBLE_SSH_RETRIES(default) = 0
ANY_ERRORS_FATAL(default) = False
BECOME_ALLOW_SAME_USER(default) = False
[...]

PR #156, opened 2020-07-09
k8s_info: Update example using vars

Akasurde:
##### SUMMARY

Fixes: #151

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/k8s_info.py

### DIFF
--- a/plugins/modules/k8s_info.py
+++ b/plugins/modules/k8s_info.py
@@ -96,6 +96,14 @@ EXAMPLES = r'''
       - app = web
       - tier in (dev, test)
 
+- name: Using vars while using label_selectors
+  community.kubernetes.k8s_info:
+    kind: Pod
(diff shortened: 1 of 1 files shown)


plugins/modules/k8s_info.py
```diff
@@ -93,8 +93,10 @@
   community.kubernetes.k8s_info:
     kind: Pod
     label_selectors:
-      - app = web
+      - "app = {{ app_label_web }}"
```
fabianvf, line 96: Would you mind adding this a as a separate example? Just want to make sure no users see the variable being set and assume it's an important piece of the `label_selectors` option

Akasurde: Make sense to me. Thanks for the review.

Issue #122, opened 2020-06-05
k8s_exec should error if the process exit code isn't 0


Yantrio:
##### SUMMARY

At the moment if you run k8s_exec and the command returns a non 0 status code the task still returns a successful change. I think it should return an error.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

k8s_exec

##### ADDITIONAL INFORMATION

This could solve issues for playbooks where commands could possibly fail, and carry on unknowningly.

```yaml
- name: Cause an exit code by using command 'false'
  community.kubernetes.k8s_exec:
    namespace: default
    pod: "mypodname"
    command: "false"

- name: This should never fire due to the error above
  .....
```

My python isn't too amazing, and I'm not familiar with the python kubernetes client, but it looks like we could utillize something similar to this https://github.com/kubernetes-client/python-base/pull/160/files in the part that parses the stream in k8s_exec here: https://github.com/ansible-collections/community.kubernetes/blob/18dfb687e6a3dbbbabec1d272f0983a8b24dfb4c/plugins/modules/k8s_exec.py#L135-L142

Due to this being released already also, this may have to be behind a property as to not break existing functionality in playbooks out there already.

I may try at implementing this and I'll put in a PR if I'm successful.

Akasurde:
@Yantrio I added `rc` in k8s_exec which returns the status code of the command, but this is changing the behavior of the module. Since we are `rc` affects ansible task status. Let me know if you want a toggle switch to enable this feature or not.

Thanks,

PR #158, opened 2020-07-10
k8s_exec: Return rc for the command executed

Akasurde:
##### SUMMARY

k8s_exec now returns command return code in output.

Fixes: #122

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
changelogs/fragments/122_k8s_exec_rc.yml
molecule/default/tasks/exec.yml
plugins/modules/k8s_exec.py

### DIFF
--- a/plugins/modules/k8s_exec.py
+++ b/plugins/modules/k8s_exec.py
@@ -30,10 +30,13 @@ requirements:
   - "openshift == 0.4.3"
   - "PyYAML >= 3.11"
 
+notes:
+- Return code C(return_code) for the command executed is added in output in version 0.11.1.
 options:
   proxy:
     description:
-    - The URL of an HTTP proxy to use for the connection. Can also be specified via K8S_AUTH_PROXY environment variable.
+    - The URL of an HTTP proxy to use for the connection.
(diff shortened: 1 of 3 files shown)


Akasurde:
@geerlingguy @fabianvf @tima This is changing behavior of the module. Module used to return True always. With this change, depending on command return code. Do you think we should add a toggle for enabling this feature?

plugins/modules/k8s_exec.py
```diff
@@ -131,15 +165,26 @@ def main():
...
             stdout.append(resp.read_stdout())
         if resp.peek_stderr():
             stderr.append(resp.read_stderr())
+    err = resp.read_channel(3)
+    err = yaml.safe_load(err)
+    if err['status'] == 'Success':
+        rc = 0
+    else:
+        rc = int(err['details']['causes'][0]['message'])
+    changed = True if rc == 0 else False
```
fabianvf, line 181: Does it make sense to assume no change on failure? Many changes could still be made by a complex command that ultimately failed

Akasurde: That's what I was thinking. I will keep the original condition of `changed` (always returning True). Thanks for confirming.

PR #160, opened 2020-07-10
k8s_cluster_info: new module

Akasurde:
##### SUMMARY

Get information about k8s APIs

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
molecule/default/converge.yml
molecule/default/tasks/api_info.yml
plugins/modules/k8s_api_info.py

### DIFF
--- a/plugins/modules/k8s_cluster_info.py
+++ b/plugins/modules/k8s_cluster_info.py
@@ -1,0 +1,247 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# Copyright (c) 2020, Abhijeet Kasurde
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = r'''
+module: k8s_cluster_info
+
+version_added: "0.11.1"
+
+short_description: Describe Kubernetes (K8s) API available and their respective versions
+
+author:
+    - Abhijeet Kasurde (@Akasurde)
+
+description:
+  - Use the OpenShift Python client to perform read operations on K8s objects.
+  - Authenticate using either a config file, certificates, password or token.
+  - Supports check mode.
+
+options:
+  invalidate_cache:
+    description:
+    - Invalidate cache before retrieving information about cluster.
+    type: bool
+    default: True
+extends_documentation_fragment:
+  - community.kubernetes.k8s_auth_options
+
+requirements:
+  - "python >= 2.7"
+  - "openshift >= 0.6"
+  - "PyYAML >= 3.11"
+'''
+
+EXAMPLES = r'''
+- name: Get Cluster information
+  community.kubernetes.k8s_cluster_info:
+  register: api_status
+
+- name: Do not invalidate cache before getting information
+  community.kubernetes.k8s_cluster_info:
+    invalidate_cache: False
+  register: api_status
+
+'''
+
+RETURN = r'''
+connection:
+  description:
+  - Connection information
+  returned: success
+  type: dict
+  contains:
+    cert_file:
+      description:
+      - Path to client certificate.
+      type: str
+      returned: success
+    host:
+      description:
+      - Host URL
+      type: str
+      returned: success
+    password:
+      description:
+      - User password
+      type: str
+      returned: success
+    proxy:
+      description:
+      - Proxy details
+      type: str
+      returned: success
+    ssl_ca_cert:
+      description:
+      - Path to CA certificate
+      type: str
+      returned: success
+    username:
+      description:
+      - Username
+      type: str
+      returned: success
+    verify_ssl:
+      description:
+      - SSL verification status
+      type: bool
+      returned: success
+version:
+  description:
+  - Information about server and client version
+  returned: success
+  type: dict
+  contains:
+    server:
+      description: Server version
+      returned: success
+      type: dict
+    client:
+      description: Client version
+      returned: success
+      type: str
+apis:
+  description:
+  - The API(s) that exists in dictionary
+  returned: success
+  type: dict
+  contains:
+    api_version:
+      description: API version
+      returned: success
+      type: str
+    categories:
+      description: API categories
+      returned: success
+      type: list
+    group_version:
+      description: Resource Group version
+      returned: success
+      type: str
+    kind:
+      description: Resource kind
+      returned: success
+      type: str
+    name:
+      description: Resource short name
+      returned: success
+      type: str
+    namespaced:
+      description: If resource is namespaced
+      returned: success
+      type: bool
+    preferred:
+      description: If resource version preferred
+      returned: success
+      type: bool
+    short_names:
+      description: Resource short names
+      returned: success
+      type: str
+    singular_name:
+      description: Resource singular name
+      returned: success
+      type: str
+    available_api_version:
+      description: All available versions of the given API
+      returned: success
+      type: list
+    preferred_api_version:
+      description: Preferred version of the given API
+      returned: success
+      type: str
+'''
+
+
+import copy
+import traceback
+
+from ansible.module_utils.basic import AnsibleModule, missing_required_lib
+from ansible.module_utils.parsing.convert_bool import boolean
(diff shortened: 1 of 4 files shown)


tima:
I agree with @fabianvf. I'd like to see cluster info included. Info about a cluster's API is just a subset of that information.

tima:
@Akasurde: In light of the suggestions here, I'd like to suggest we rename this module `k8s_cluster_info`.

plugins/modules/k8s_api_info.py
```diff
@@ -0,0 +1,108 @@
...
+                'available_api_versions': [],
+            }
+            for v in api.versions:
+                if v.version == api.preferred_version.version:
+                    current_api['preferred_api_version'] = v.version
+                current_api['available_api_versions'].append(v.version)
+            results.append(current_api)
+        results.append({
+            'api_name': 'core',
+            'available_api_versions': client.CoreApi().get_api_versions().versions,
```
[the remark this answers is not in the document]
Akasurde: @fabianvf Do you know a way with which we can check this?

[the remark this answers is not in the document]
fabianvf:
```
client = self.get_client()
resources = list(client.resources)
```
That'll give you a list of lists of resource objects, like:

```
[[<Resource(v1/bindings)>],                                                                                                                                                                                                                   
 [<ResourceList(v1/bindings)>],                                                                                                                                                                                                               
 [<Resource(v1/componentstatuses)>],                                                                                                                                                                                                          
 [<ResourceList(v1/componentstatuses)>],                                                                                                                                                                                                      
 [<Resource(v1/configmaps)>],                                                                                                                                                                                                                 
 [<ResourceList(v1/configmaps)>],                                                                                                                                                                                                             
 [<Resource(v1/endpoints)>],                                                                                                                                                                                                                  
 [<ResourceList(v1/endpoints)>],                                                                                                                                                                                                              
 [<Resource(v1/events)>],                                                                                                                                                                                                                     
 [<ResourceList(v1/events)>],                                                                                                                                                                                                                 
 [<Resource(v1/limitranges)>],                                                                                                                                                                                                                
 [<ResourceList(v1/limitranges)>],                                                                                                                                                                                                            
 [<Resource(v1/namespaces)>],                                                                                                                                                                                                                 
 [<ResourceList(v1/namespaces)>],                                                                                                                                                                                                             
 [<Resource(v1/nodes)>],                                                                                                                                                                                                                      
 [<ResourceList(v1/nodes)>],
...]
```
Where each `Resource` object contains the relevant information for constructing the response:

```python
$ list(client.resources)[0][0].__dict__
{'prefix': 'api',
 'group': '',
 'api_version': 'v1',
 'kind': 'Binding',
 'namespaced': True,
 'verbs': ['create'],
 'name': 'bindings',
 'preferred': True,
 'client': <openshift.dynamic.client.DynamicClient at 0x7f7a1f2bd550>,
 'singular_name': 'binding',
 'short_names': None,
 'categories': None,
 'subresources': {},
 'extra_args': {}}
```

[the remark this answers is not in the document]
fabianvf: The `ResourceList` objects can be ignored, they don't actually have a representation on the API side

plugins/modules/k8s_cluster_info.py
```diff
@@ -0,0 +1,185 @@
...
+'''
+
+EXAMPLES = r'''
+- name: Get an existing Service object
+  community.kubernetes.k8s_cluster_info:
+  register: api_status
+'''
+
+RETURN = r'''
+apis:
```
fabianvf, line 41: We should return some information in addition to the APIs. 

I'd like to see some information related to the `DynamicClient.configuration` object (an instance of the [kubernetes.client.Configuration](https://github.com/kubernetes-client/python/blob/master/kubernetes/client/configuration.py#L25)). We could pull the `host`, `api_key`, `username`, `password`, `verify_ssl`, `ssl_ca_cert`, `cert_file`, `key_file` and `proxy` (many of those would need to be `no_log`). This will allow the user to determine how the client is interacting with the cluster, even when the client was loaded via incluster config or kubeconfig (I've run into this need in the `okd` collection tests already)

I'd also like to see information about the cluster version (`DynamicClient.version`), and possibly about the client version as well (`openshift.__version__`), similarly to how `kubectl version` works.

Final output structure could look something like:

```
apis:
  # everything you have already
connection:
  host: 
  api_key:
  username:
  password:
  ...
version:
  server:
  client:
```

plugins/modules/k8s_cluster_info.py
```diff
@@ -0,0 +1,185 @@
...
+  description:
+  - The API(s) that exists in dictionary
+  returned: success
+  type: dict
+  contains:
+    api_version:
```
fabianvf, line 47: I would nix `group_version` and make `api_version` `DynamicClient.group_version`. It's a bit confusing, but to users `apiVersion` means `{group}/{api_version}`, but internally most clients deal with them separately.

plugins/modules/k8s_cluster_info.py
```diff
@@ -0,0 +1,185 @@
...
+      type: str
+    namespaced:
+      description: If resource is namespaced
+      returned: success
+      type: bool
+    preferred:
+      description: If resource version preferred
+      returned: success
+      type: bool
+    prefix:
```
fabianvf, line 75: This is more internal, I don't think it needs to be exposed

plugins/modules/k8s_cluster_info.py
```diff
@@ -0,0 +1,185 @@
...
+      type: str
+    short_names:
+      description: Resource short names
+      returned: success
+      type: str
+    singular_name:
+      description: Resource singluar name
+      returned: success
+      type: str
+    urls:
```
fabianvf, line 87: This is also internal, if someone wants to call these URLs directly they should probably be able to form them by hand

plugins/modules/k8s_cluster_info.py
```diff
@@ -0,0 +1,185 @@
...
+                'name': resource.name,
+                'namespaced': resource.namespaced,
+                'preferred': resource.preferred,
+                'prefix': resource.prefix,
+                'short_names': resource.short_names if resource.short_names else [],
+                'singular_name': resource.singular_name,
+                'urls': resource.urls,
+            }
+
+        for api in client.ApisApi().get_api_versions().groups:
```
fabianvf, line 164: we can probably skip this, we already store whether a resource is `preferred` or not which should be enough to filter on if the information is needed

plugins/modules/k8s_cluster_info.py
```diff
@@ -0,0 +1,185 @@
...
+        if not HAS_K8S_INSTANCE_HELPER:
+            self.module.fail_json(msg=missing_required_lib("openshift >= 0.6.2", reason="for merge_type"),
+                                  exception=k8s_import_exception)
+
+        super(KubernetesInfoModule, self).__init__()
+
+    def execute_module(self):
+        self.client = self.get_api_client()
+        results = {}
+        for resource in list(self.client.resources):
```
fabianvf, line 146: We may want to expose the option to invalidate the cache (or invalidate it by default), as it's possible that APIs are added/removed without getting picked up (internally when you attempt to contact a specific API we will update properly, but when iterating over the APIs it may not pick up CRD changes to an existing API group). You can invalidate the cache with `self.client.resources.invalidate_cache()`

fabianvf (changes requested):
a few comments on what information we can/should surface to the user, plus a possible issue with caching, other than that :+1:

fabianvf:
Also we should link it in the README

plugins/modules/k8s_cluster_info.py
Akasurde: Done.

plugins/modules/k8s_cluster_info.py
Akasurde: Done.

plugins/modules/k8s_cluster_info.py
Akasurde: Done.

fabianvf (approved):
Looks great to me, we could add some tests for the returned `connection`/`version` fields but other than that :+1: (and we could do it in a followup as well)

PR #180, opened 2020-08-04
raw: Handle exception raised by underlying APIs

Akasurde:
##### SUMMARY

kind can be provided by number of ways. It is diffcult to
find all cases. So, handle exceptions raised by underlying
APIs and provide good error message to user.

Fixes: ansible/ansible#63652

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/module_utils/raw.py

### DIFF
--- a/plugins/module_utils/raw.py
+++ b/plugins/module_utils/raw.py
@@ -27,6 +27,7 @@ import traceback
 from ansible.module_utils.basic import missing_required_lib
 from ansible_collections.community.kubernetes.plugins.module_utils.common import AUTH_ARG_SPEC, COMMON_ARG_SPEC
 from ansible.module_utils.six import string_types
+from ansible.module_utils._text import to_native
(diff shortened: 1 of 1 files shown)


plugins/module_utils/raw.py
```diff
@@ -175,7 +176,7 @@ def execute_module(self):
                 continue
             kind = definition.get('kind', self.kind)
             api_version = definition.get('apiVersion', self.api_version)
-            if kind.endswith('List'):
+            if kind and kind.endswith('List'):
```
fabianvf, line 179: I think if kind isn't defined here we're broken anyway, this should have been validated earlier. We could add an additional failout here if either kind or api_version aren't defined

tima:
@geerlingguy and @willthames -- any thoughts here?

willthames:
Agree with @fabianvf's comments but this is a good change without those revisions, they would just improve it further (so not a blocker for me)

PR #196, opened 2020-08-18
Issue #53: Add probot/stale bot configuration to ID and close stale issues.

### LINKED ISSUE #53: Add Probot/stale bot to mark issues as stale and close after certain period
##### SUMMARY

For my own GitHub repos, I've [configured the probot/stale bot](https://www.jeffgeerling.com/blog/2020/enabling-stale-issue-bot-on-my-github-repositories) to mark issues as stale after 90 days with no activity, and close after an additional 30 days if no further activity is found.

For the long-term maintenance of this collection, it might be nice to do the same thing here, just to make sure that 'rotting' issues and PRs (especially ones with no active interest) are automatically pruned.

Not sure the exact time we'd want to configure, but it's pretty easy to set up via the stale.yml configuration...

##### ISSUE TYPE

- Feature Idea

##### COMPONENT NAME

Organization / maintenance

##### ADDITIONAL INFORMATION

N/A

geerlingguy:
Closes #53.﻿

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,60 @@
+---
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 90
+
+# Number of days of inactivity before an Issue or Pull Request with the stale
+# label is closed. Set to false to disable. If disabled, issues still need to be
+# closed manually, but will remain marked as stale.
+daysUntilClose: 30
+
+# Only issues or pull requests with all of these labels are check if stale.
+# Defaults to `[]` (disabled)
+onlyLabels: []
+
+# Issues or Pull Requests with these labels will never be considered stale. Set
+# to `[]` to disable
+exemptLabels:
+  - pinned
+  - security
+  - planned
+  - priority/critical
+  - verified
+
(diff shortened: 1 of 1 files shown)


tima:
LGTM also.

@geerlingguy Do we need to add any additional labels or will the probot handle label creation?

geerlingguy:
@tima - If we want to add certain new labels to mark issues as 'not stale permanently', we would need to add those labels to the bot config. Right now it's just:

```yaml
exemptLabels:
   - pinned
   - security
   - planned
   - priority/critical
   - verified
```

tima:
@geerlingguy 

>    - priority/critical
>    - verified

OK. We have these already.

>    - security
>    - planned

These could be useful. I'll add these. 

>    - pinned

Not sure what we'd use this one for.

fabianvf:
@tima I think pinned would likely be equivalent to `lifecycle/frozen`, which generally means "This is a real issue, it should not be closed. It is not high priority (or can't be worked on yet), and it is not currently on the list of planned work." Basically an issue that could be sitting around for a few months to a year or more, but it would be bad if it was forgotten

geerlingguy:
Ah, in that case I'll add that as another tag to use for ignoring stale.

PR #199, opened 2020-08-20
k8s_*: Group argument_spec accroding to usage

### LINKED ISSUE #36: Remove module inheritance
##### SUMMARY

Currently we have the [`KubernetesAnsibleModule`](https://github.com/ansible-collections/kubernetes/blob/master/plugins/module_utils/common.py#L267) and [`KubernetesRawModule`](https://github.com/ansible-collections/kubernetes/blob/master/plugins/module_utils/raw.py#L65) that most of our modules inherit from, but this is a bad pattern and has already bitten us with the `k8s_scale` module. We have an [`AnsibleMixin`](https://github.com/ansible-collections/kubernetes/blob/master/plugins/module_utils/common.py#L147) that works fine, though we should remove the argspec from it.

The main reason this pattern is harmful is that parameters are shared across modules, and adding or changing arguments in one will passively propagate to subclasses of that module. `k8s_scale` for example, inherits parameters from `k8s`, and has broken in the past when we added arguments to the `k8s` module ([issue here](https://github.com/ansible/ansible/issues/59021)). Each module should own the arguments it accepts, and code sharing should be limited to shared utilities rather than module definitions.

##### ISSUE TYPE
- Technical Debt

Akasurde:
##### SUMMARY

Partially fix #36

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/module_utils/scale.py

### DIFF
--- a/plugins/module_utils/scale.py
+++ b/plugins/module_utils/scale.py
@@ -21,7 +21,7 @@ __metaclass__ = type
 
 import copy
 
-from ansible_collections.community.kubernetes.plugins.module_utils.common import AUTH_ARG_SPEC, COMMON_ARG_SPEC
+from ansible_collections.community.kubernetes.plugins.module_utils.common import AUTH_ARG_SPEC, RESOURCE_ARG_SPEC, NAME_ARG_SPEC
 from ansible_collections.community.kubernetes.plugins.module_utils.common import KubernetesAnsibleModule
 from ansible.module_utils.six import string_types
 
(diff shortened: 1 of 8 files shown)


plugins/module_utils/scale.py
```diff
@@ -33,6 +33,20 @@
...
 
 SCALE_ARG_SPEC = {
+    'resource_definition': {
+        'type': list_dict_str,
+        'aliases': ['definition', 'inline']
+    },
+    'src': {
+        'type': 'path',
+    },
+    'kind': {},
```
fabianvf, line 43: would it make sense to group together the common parameters into logical pieces that various argspecs could make use of (and each group would have a corresponding doc fragement)? ie, `api_version`, `kind`, `name` and `namespace`, `resource_definition` and `src`, `wait`, `wait_timeout` and `wait_condition`. They wouldn't be inheriting argspecs from another module, just composing them from sets of shared parameters that ideally can just be updated in one place

fabianvf: basically the same as we do with `AUTH_ARG_SPEC`

plugins/modules/k8s_log.py
```diff
@@ -26,36 +26,8 @@
...
-    - api
-    - version
-    type: str
-  kind:
-    description:
-    - Use to specify an object model. Use in conjunction with I(api_version), I(name), and I(namespace) to identify a
-      specific object.
-    - If using I(label_selector), cannot be overridden
-    required: no
-    default: Pod
```
fabianvf, line 46: Hmm, this one has a different default than the other modules, that's a bit inconvenient

Issue #106, opened 2020-05-27
Implement helm install --replace


T1loc:
##### SUMMARY

Implement helm install --replace

#### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

helm

##### ADDITIONAL INFORMATION

When a release is uninstalled with `helm uninstall --keep-history`, we are not able to perform a `helm upgrade -i`

```yaml
helm list --all -o yaml                                                                                                                                                                                           
- app_version: ""
  chart: xxx-xxx
  name: test
  namespace: florian
  revision: "1"
  status: uninstalled
  updated: 2020-05-27 20:07:15.487307 +0200 CEST
```

To install by reusing the release name and then keep the history
```yaml
helm install --replace --namespace=florian test test 
```

Open Questions : 
- Should we add a replace option (True/False) only available when the release status is uninstalled ?
- Any other way to answer the problem ?

PR #214, opened 2020-08-31
helm: Add replace parameter

Akasurde:
##### SUMMARY

Fixes: #106

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
changelogs/fragments/106-helm_replace.yml
molecule/default/roles/helm/tasks/tests_chart.yml
plugins/modules/helm.py

### DIFF
--- a/plugins/modules/helm.py
+++ b/plugins/modules/helm.py
@@ -115,6 +115,13 @@ options:
     type: bool
     default: False
     version_added: "0.11.1"
+  replace:
+    description:
+      - Reuse the given name, only if that name is a deleted release which remains in the history.
+      - This is unsafe in production environment.
+    type: bool
+    default: False
+    version_added: "1.11.0"
 extends_documentation_fragment:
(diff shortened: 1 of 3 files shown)


Akasurde:
@LucasBoisserie Could you please review this? Thanks.

plugins/modules/helm.py
```diff
@@ -455,6 +473,10 @@ def main():
     # keep helm_cmd_common for get_release_status in module_exit_json
     helm_cmd = helm_cmd_common
     if release_state == "absent" and release_status is not None:
+
+        if not replace and replace:
```
fabianvf, line 477: I think one of these variables is probably supposed to be different

fabianvf (commented):
one issue I spotted, otherwise looks great

plugins/modules/helm.py
Akasurde: oh My bad. I will update.

Issue #18, opened 2020-02-11
Add `k8s_wait` module


geerlingguy:
##### SUMMARY

Just wanted to throw this out here, since it's something that would be convenient in a number of circumstances.

A common pattern for K8s deployments is:

  1. Apply a manifest to create a new Deployment.
  2. Wait for all the Pods in this Deployment to be Ready.
  3. Do other stuff.

Currently, for number 2, you can futz around with the returned data from `k8s` or `k8s_info` and use until/retries to get something working, or you can do a more simple method (if you have `kubectl` available) using `kubectl wait`:

    kubectl wait --for=condition=Ready pods --selector app=my-app-name --timeout=60s

At a basic level, I'd want something like:

```yaml
- name: Wait for my-app-name pods to be ready.
  k8s_wait:
    for: condition=Ready
    type: pods
    selector:
      - app=my-app-name
    timeout: 60s
```

Something along those lines... not sure. But it would be nice to be able to specify this in a more structured way, and not have to rely on `kubectl` being present for a task like:

```yaml
- name: Wait for my-app-name pods to be ready.
  command: >
    kubectl wait --for=condition=Ready
    pods --selector app=my-app-name --timeout=60s
  changed_when: false
```

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

N/A

##### ADDITIONAL INFORMATION

N/A

geerlingguy:
This feature request stems from https://github.com/geerlingguy/ansible-collection-k8s/issues/5, where I was originally ruminating over it.

fabianvf:
@geerlingguy does the `wait` parameter in the `k8s` module fall short?

geerlingguy:
To be honest, I might have not been thinking about that parameter when I originally wrote up the issue in my collection repo... I think it was when I was writing up this example in Ansible for Kubernetes: https://github.com/geerlingguy/ansible-for-kubernetes/blob/master/cluster-local-vms/test-deployment.yml#L15-L27

I'll see if I can get that to work with the `wait` parameter instead...

geerlingguy:
It works great, not sure why I didn't use the `wait` parameter earlier. Downstream issue: https://github.com/geerlingguy/ansible-for-kubernetes/issues/32

Closing this request as it's redundant (and would just add complexity where none is needed).

bergmannf:
I think a module like this might still be useful.

[...]

PR #216, opened 2020-09-02
k8s_info: Add wait functionality

Akasurde:
##### SUMMARY

User can specify wait in k8s_info module.

Fixes: #18

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/module_utils/raw.py
plugins/modules/k8s.py
plugins/modules/k8s_info.py

### DIFF
--- a/plugins/module_utils/common.py
+++ b/plugins/module_utils/common.py
@@ -142,6 +142,21 @@ AUTH_ARG_SPEC = {
     },
 }
 
+WAIT_ARG_SPEC = dict(
+    wait=dict(type='bool', default=False),
+    wait_sleep=dict(type='int', default=5),
+    wait_timeout=dict(type='int', default=120),
+    wait_condition=dict(
+        type='dict',
+        default=None,
+        options=dict(
+            type=dict(),
+            status=dict(default=True, choices=[True, False, "Unknown"]),
+            reason=dict()
+        )
+    )
+)
+
 # Map kubernetes-client parameters to ansible parameters
 AUTH_ARG_MAP = {
     'kubeconfig': 'kubeconfig',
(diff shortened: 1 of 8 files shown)


plugins/modules/k8s_info.py
```diff
@@ -42,10 +42,21 @@
     description: List of field selectors to use to filter results
     type: list
     elements: str
+  state:
+    description:
+    - Desired state of the object to gather information from.
+    - Only applicable if C(wait), C(wait_sleep), C(wait_condition), and C(wait_timeout) specified.
+    - If you are waiting for object to be created by Kubernetes, set this to C(present).
+    - If you are waiting for object to be deleted by Kubernetes, set this to C(absent).
+    - This parameter is different from regular Ansible state parameter.
```
fabianvf, line 51: Is there a convention for this or does `state` still make the most sense?

willthames:
Do we have a use case for this change? I'm just wondering in what circumstances you'd want to wait in k8s_info rather than just waiting in k8s, or using a retry loop around k8s_info

geerlingguy:
@willthames specifically see: https://github.com/ansible-collections/community.kubernetes/issues/18#issuecomment-682153809

willthames:
thanks @geerlingguy. Do we ever think we'll care about waiting for things to be deleted? (I'm just keen to avoid adding `state` explicitly)

plugins/modules/k8s_info.py
willthames: I'd prefer not to add it, but have `wait_state` if we really need it

fabianvf: yeah, I think the `absent` behavior can actually be achieved like so:

```
k8s_info:
  api_version: v1
  kind: Namespace
  name: testing
register: result
until: not result.resources
```

so really no need for `state` here IMO

Akasurde: I removed `state`.

Akasurde:
@willthames @fabianvf @geerlingguy I removed the `state` parameter and added test. Please review once. Thanks.

plugins/module_utils/common.py
```diff
@@ -318,7 +355,9 @@ def _wait_for_elapsed():
         response = None
         while _wait_for_elapsed() < timeout:
             try:
-                response = resource.get(name=name, namespace=namespace)
+                response = resource.get(name=name, namespace=namespace,
```
fabianvf, line 358: One thing to note, I think previously the implementation of `wait` dealt only with a single resource, but multiple resources can be retrieved via a selector. We either need to iterate over the returned resources here or only pass individual resource names to the `wait` function.

fabianvf: We could also implement a decorator for the predicates that deals with iterating over `ResourceList` objects if they're passed in. Any way works as long as we just iterate over the individual resources at some point.

fabianvf: We should also add a test for the `k8s_info` wait that handles multiple objects matching the selector (ie, two pods that share a label)

fabianvf (changes requested):
One spot I noticed a potential issue

PR #217, opened 2020-09-02
Refactor inventory plugin

fabianvf:
##### SUMMARY
Inventory plugin only adds valid hosts and groups
    
- No longer adds services to lists of hosts
- Group names are now sanitized
- Removed the `*_pods` groups, as pods are the only valid hosts anyway
- Add new groups for Deployments, Daemonsets and StatefulSets, allowing
  easy access to pods that are created by those resources
- Move extract_selector method to common

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
plugins/inventory/k8s.py

##### ADDITIONAL INFORMATION
Technically modifying the group names to be valid could be considered a breaking change, though it's probably a bug that they weren't in the first place.

### DIFF
--- a/plugins/inventory/k8s.py
+++ b/plugins/inventory/k8s.py
@@ -154,6 +154,12 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable, K8sAnsibleM
     connection_plugin = "community.kubernetes.kubectl"
     transport = "kubectl"
 
+    PARENT_RESOURCES = [
+        {'api_version': 'apps/v1', 'kind': 'Deployment'},
+        {'api_version': 'apps/v1', 'kind': 'DaemonSet'},
+        {'api_version': 'apps/v1', 'kind': 'StatefulSet'},
+    ]
+
     def parse(self, inventory, loader, path, cache=True):
         super(InventoryModule, self).parse(inventory, loader, path)
         cache_key = self._get_cache_prefix(path)
@@ -238,10 +244,17 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable, K8sAnsibleM
 
     def get_pods_from_parents(self, client, name, namespace, namespace_group):
 
+<<<<<<< HEAD
         v1_pods = client.resources.get(api_version="v1", kind="Pod")
         for kind in ["Deployment", "Daemonset", "StatefulSet"]:
             try:
                 resource = client.resources.get(api_version="apps/v1", kind=kind)
+=======
+        v1_pods = client.resources.get(api_version='v1', kind='Pod')
+        for item in self.PARENT_RESOURCES:
+            try:
+                resource = client.resources.get(api_version=item['api_version'], kind=item['kind'])
+>>>>>>> Make parent resources more easily configurable
                 instances = resource.get(namespace=namespace)
             except Exception:
                 # TODO Could be expected due to RBAC or odd cluster, maybe should log a warning or something?
@@ -263,6 +276,7 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable, K8sAnsibleM
                     )
                 except DynamicApiError as exc:
                     self.display.debug(exc)
+<<<<<<< HEAD
                     raise K8sInventoryException(
                         "Error fetching Pod list: %s" % format_dynamic_api_exc(exc)
                     )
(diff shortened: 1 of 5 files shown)


plugins/inventory/k8s.py
```diff
@@ -171,35 +177,43 @@ def setup(self, config_data, cache, cache_key):
...
+
+                self.get_pods_for_namespace(client, name, namespace, namespace_group)
+                self.get_pods_from_parents(client, name, namespace, namespace_group)
 
     @staticmethod
     def get_default_host_name(host):
         return host.replace('https://', '').replace('http://', '').replace('.', '-').replace(':', '_')
 
+    @staticmethod
+    def sanitize(name):
```
Akasurde, line 212: There is a function called [`_sanitize_group_name`](https://github.com/ansible/ansible/blob/48f12c14e9ff2b04c49cf6f90a65f4a9ecbb7fc7/lib/ansible/plugins/inventory/__init__.py#L153). Can we use this?

fabianvf: awesome, yeah we can definitely use that

geerlingguy:
I like:

  * No longer adds services to lists of hosts
  * Removed the `*_pods` groups, as pods are the only valid hosts anyway

And I have never used those features (and have only tested the inventory plugin once anyways), but if we merge those changes, we will need to bump to a 2.0.0 version.

We could merge the rest of the changes in 1.x, I think—and then maybe you can move those other two changes to a PR that we could put in a 2.0.0 milestone?

tima:
@fabianvf I agree about the group renaming and have to wonder if many at all where using the parts you touched. A breaking change is a breaking change that I'm going to attach this one to the 2.0 project.

plugins/inventory/k8s.py
```diff
@@ -209,154 +223,105 @@ def get_available_namespaces(self, client):
...
-    def get_pods_for_namespace(self, client, name, namespace):
+    def get_pods_from_parents(self, client, name, namespace, namespace_group):
+
+        v1_pods = client.resources.get(api_version='v1', kind='Pod')
+        for item in self.PARENT_RESOURCES:
+            try:
+                resource = client.resources.get(api_version=item['api_version'], kind=item['kind'])
+                instances = resource.get(namespace=namespace)
+            except Exception:
+                # TODO Could be expected due to RBAC or odd cluster, maybe should log a warning or something?
```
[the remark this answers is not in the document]
geerlingguy: That I do not; you'd have to get @Akasurde's input, I have never built an inventory plugin (yet—I'm going to work on getting one merged for DigitalOcean soon).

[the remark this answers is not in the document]
Akasurde: We can use `strict` parameter and depend upon strictness (`strict: true`) we take action.

like -

```
if strict:
   raise AnsibleError("Could not...
else:
   display.warning("Could not...
```

You can find the example https://github.com/ansible-collections/vmware/blob/341c3375941d3119a61c10359c85a4caa8082e18/plugins/inventory/vmware_vm_inventory.py#L446 and https://github.com/ansible-collections/vmware/blob/341c3375941d3119a61c10359c85a4caa8082e18/plugins/inventory/vmware_vm_inventory.py#L807

plugins/inventory/k8s.py
```diff
@@ -14,7 +14,7 @@
     short_description: Kubernetes (K8s) inventory source
 
     description:
-      - Fetch containers and services for one or more clusters
+      - Fetch containers in pods for one or more clusters
```
maxamillion, line 17: Is this a change in functionality or is this a correction to the documentation?

geerlingguy: That's a change in functionality; this PR would limit the scope of the inventory plugin to something that's more practical in real-world use, but because it changes the behavior and removes a feature, that should probably be put into a 2.x.

Akasurde:
recheck

plugins/module_utils/common.py
```diff
@@ -782,6 +782,42 @@ def create_project_request(self, definition):
         result['method'] = 'create'
         return result
 
+    def extract_selectors(self, instance):
```
goneri, line 785: The method don't use `self`, can we convert it in a function and add test-coverage?

goneri (approved):
Looks good to me, but I still need to test the code.

tima:
I don't feel we have enough feedback on the validity or necessity of this plugin. I'd like to see us hold up on this until we get that for or against it's continued inclusion.

PR #227, opened 2020-09-09
Replace KubernetesAnsibleModule class with dummy class

Akasurde:
##### SUMMARY

* Make a AnsibleMixin parent class for every module
* Remove KubernetesAnsibleModule class
* Modified k8s_log

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/k8s_log.py

### DIFF
--- a/plugins/module_utils/common.py
+++ b/plugins/module_utils/common.py
@@ -24,7 +24,7 @@ import os
 import traceback
 
 
-from ansible.module_utils.basic import missing_required_lib
+from ansible.module_utils.basic import AnsibleModule, missing_required_lib
 from ansible.module_utils.six import iteritems, string_types
 from ansible.module_utils._text import to_native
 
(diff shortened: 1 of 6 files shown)


geerlingguy (commented):
The structure of the change looks okay to me, though the change seems to have introduced a cascade of test failures. Assuming we resolve those I'm fine with this change.

I haven't taken a peek, but do you know if anything in the Kubevirt modules in `community.general` would blow up after making these changes; I noticed they import from `KubernetesRawModule`.

Akasurde:
> The structure of the change looks okay to me, though the change seems to have introduced a cascade of test failures. Assuming we resolve those I'm fine with this change.
> 

Yes, Since we removed KubernetesAnsibleModule from only one module, the rest are in pipeline.

> I haven't taken a peek, but do you know if anything in the Kubevirt modules in `community.general` would blow up after making these changes; I noticed they import from `KubernetesRawModule`.

Good. Thanks for reminding me that. I will take care of that as well once this is resolved.

geerlingguy (approved):
LGTM, but I would like to make sure at least one or two other maintainers deeper into any other module integrations (I know of kubevirt that might need adjustments, but I don't know if anyone else might be relying on `KubernetesAnsibleModule`...?).

The other question: would this be something we can merge into a 1.1, or do we need to go to 2.0.0 with this change?

maxamillion (approved):
LGTM :+1:

tima:
@fabianvf could you weigh in on a dummy class here to avoid breaking anyone.

plugins/module_utils/common.py
```diff
@@ -265,28 +274,6 @@ def diff_objects(existing, new):
             result['after'] = diff[1]
         return not diff, result
 
-
-class KubernetesAnsibleModule(AnsibleModule, K8sAnsibleMixin):
```
fabianvf, line 269: We can probably leave this and the `KubernetesRawModule` classes around as dummies with warnings to avoid a breaking API change

fabianvf (changes requested):
Sorry that last review should have been marked as `Request changes`

fabianvf (approved):

felixfontein:
I think this PR breaks the unit tests in community.general (see f.ex. https://app.shippable.com/github/ansible-collections/community.general/runs/4278/15/tests). Could someone please check whether this also affects the kubevirt modules?

Akasurde:
@felixfontein I have raised a PR to fix failing kubevirt tests.

Akasurde:
Community.general - https://github.com/ansible-collections/community.general/pull/1070

PR #228, opened 2020-09-14
Do not mark task as changed when diff is irrelevant

fabianvf:
##### SUMMARY

When the diff contains changes only to the fields `metadata.generation`
or `metadata.resourceVersion`, do not mark the task as changed. Instead,
   emit a warning highlighting that the API itself may not be
   idempotent, but that there was no meaningful difference between the
   desired and actual state of the resource.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

plugins/modules/k8s.py

##### ADDITIONAL INFORMATION

```paste below

```


maxamillion (approved):
LGTM :+1:

PR #230, opened 2020-09-16
k8s: add support for template

Akasurde:
##### SUMMARY

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/action/k8s_info.py
plugins/module_utils/raw.py

### DIFF
--- a/molecule/default/tasks/template.yml
+++ b/molecule/default/tasks/template.yml
@@ -1,0 +1,155 @@
+---
+- block:
+    - set_fact:
+        template_namespace: template-test
+
+    - name: Ensure namespace exists
+      k8s:
+        definition:
+          apiVersion: v1
+          kind: Namespace
+          metadata:
+            name: "{{ template_namespace }}"
+
+    - name: Check if k8s_service does not inherit parameter
+      community.kubernetes.k8s_service:
+        template: "pod_template_one.j2"
+        state: present
+      ignore_errors: yes
+      register: r
+
+    - name: Check for expected failures in last tasks
+      assert:
+        that:
+          - r.failed
+          - "'is only supported parameter for' in r.msg"
+
+    - name: Specify both definition and template
+      community.kubernetes.k8s:
+        state: present
+        template: "pod_template_one.j2"
+        definition:
+          apiVersion: apps/v1
+          kind: Deployment
+          metadata:
+            name: apply-deploy
+            namespace: "{{ template_namespace }}"
+          spec:
+            replicas: 1
+            selector:
+              matchLabels:
+                app: "{{ k8s_pod_name }}"
+      vars:
+        k8s_pod_name: pod
+        k8s_pod_namespace: "{{ template_namespace }}"
+      register: r
+      ignore_errors: yes
+
+    - name: Check if definition and template are mutually exclusive
+      assert:
+        that:
+          - r.failed
+          - "'parameters are mutually exclusive' in r.msg"
+
+    - name: Specify both src and template
+      community.kubernetes.k8s:
+        state: present
+        src: "../templates/pod_template_one.j2"
+        template: "pod_template_one.j2"
+      vars:
+        k8s_pod_name: pod
+        k8s_pod_namespace: "{{ template_namespace }}"
+      register: r
+      ignore_errors: yes
+
+    - name: Check if src and template are mutually exclusive
+      assert:
+        that:
+          - r.failed
+          - "'parameters are mutually exclusive' in r.msg"
+
+    - name: Create pod using template (direct specification)
+      community.kubernetes.k8s:
+        template: "pod_template_one.j2"
+        wait: yes
+      vars:
+        k8s_pod_name: pod-1
+        k8s_pod_namespace: "{{ template_namespace }}"
+      register: r
+
+    - name: Assert that pod creation succeeded using template
+      assert:
+        that:
+          - r is successful
+
+    - name: Create pod using template with wrong parameter
+      community.kubernetes.k8s:
+        template:
+          - default
+        wait: yes
+      vars:
+        k8s_pod_name: pod-2
+        k8s_pod_namespace: "{{ template_namespace }}"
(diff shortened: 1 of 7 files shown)


plugins/action/k8s_info.py
```diff
@@ -71,6 +71,42 @@ def run(self, tmp=None, task_vars=None):
         if src:
             new_module_args['src'] = src
 
+        template = self._task.args.get('template', None)
```
geerlingguy, line 74: The new param needs docs—and I would suggest also making sure to put something along the lines of "the template must be local to the machine running the Ansible playbook" or something like that. I know one of the original objections to template support was that you couldn't use templates on a remote source.

geerlingguy (changes requested):

maxamillion (approved):
LGTM :+1:

molecule/default/tasks/template.yml
```diff
@@ -0,0 +1,154 @@
...
+      register: r
+      ignore_errors: True
+
+    - name: Assert that pod creation failed using template
+      assert:
+        that:
+          - r is failed
+          - "'Error while reading template file' in r.msg"
+
+    - name: Create pod using template
```
geerlingguy, line 102: Can you change desc here to "Create pod using template (path parameter)"

molecule/default/tasks/template.yml
```diff
@@ -0,0 +1,154 @@
...
+      register: r
+      ignore_errors: yes
+
+    - name: Check if src and template are mutually exclusive
+      assert:
+        that:
+          - r.failed
+          - "'parameters are mutually exclusive' in r.msg"
+
+    - name: Create pod using template
```
geerlingguy, line 71: Can you change desc here to "Create pod using template (direct specification)"

molecule/default/tasks/template.yml
```diff
@@ -0,0 +1,154 @@
...
+        k8s_pod_name: pod-3
+        k8s_pod_namespace: "{{ template_namespace }}"
+      register: r
+
+    - name: Assert that pod creation succeeded using template
+      assert:
+        that:
+          - r is successful
+
+    - name: Create pod using template
```
geerlingguy, line 117: Can you change desc here to "Create pod using template (different variable string)"

geerlingguy (changes requested):
Just for easier browsing of test results, I suggested a few lines to change in the test tasks.

molecule/default/tasks/template.yml
```diff
@@ -0,0 +1,154 @@
...
+      k8s:
+        definition:
+          apiVersion: v1
+          kind: Namespace
+          metadata:
+            name: "{{ template_namespace }}"
+
+    - name: Check if k8s_service does not inherit parameter
+      community.kubernetes.k8s_service:
+        template: "../templates/pod_template_one.j2"
```
fabianvf, line 16: Will this automatically look in the `templates` directory as the `template` module would? If so, could we test that?

plugins/action/k8s_info.py
```diff
@@ -71,6 +75,118 @@ def run(self, tmp=None, task_vars=None):
...
+        if template:
+            # template is only supported by k8s module.
+            if self._task.action not in ('k8s', 'community.kubernetes.k8s'):
```
fabianvf, line 81: Would this have any impact on us using this feature in `community.okd.k8s`? Or would we just need to write our own action plugin there (if possible it would probably be nicer to reuse this one, as there probably wouldn't be any differences in functionality). Would it be possible to instead check the argspec to see if it includes a template parameter or something?

plugins/action/k8s_info.py
```diff
@@ -71,6 +75,118 @@ def run(self, tmp=None, task_vars=None):
...
+                raise AnsibleActionFail("'template' is only supported parameter for 'k8s' module.")
+            # resource_definition = self._task.args.get('resource_definition')
```
fabianvf, line 83: Looks like there's some dead code

plugins/action/k8s_info.py
Akasurde: I will include `community.okd.k8s` in this list so that the action plugin works with okd collection.

Akasurde: action plugin is executed before the argspec check in the module, so it better to check here.

molecule/default/tasks/template.yml
Akasurde: Yes, templates under `templates` directory will be detected.

plugins/action/k8s_info.py
```diff
@@ -71,6 +75,118 @@ def run(self, tmp=None, task_vars=None):
         if src:
             new_module_args['src'] = src
 
+        template = self._task.args.get('template', None)
+        if template:
+            # template is only supported by k8s module.
+            if self._task.action not in ('k8s', 'community.kubernetes.k8s', 'community.okd.k8s'):
```
fabianvf, line 81: Hmm, I still am not a fan of naming specific modules here (and referencing a module in another collection seems odd), is there not a way to do something like verify the `template` parameter is in the argspec of the module?

maxamillion: @fabianvf that's a fair point, thank you for raising it :+1:

Akasurde: @fabianvf I am afraid that there is no way to do that. Since the Action plugin runs before argspec validation which is in the module code. Ansible core repos also do this kind of validation https://github.com/ansible/ansible/pull/71824/files#diff-b77962b6b54a830ec373de0602918318R86 (referring modules from other collections).

fabianvf: hmm, if there's not an easy way to do it in a more generic way then I'm ok with merging as-is

fabianvf (approved):

plugins/action/k8s_info.py
Akasurde: OK.

PR #235, opened 2020-09-23
k8s_info: Add support for wait

Akasurde:
##### SUMMARY

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
changelogs/fragments/18_k8s_info_wait.yml
molecule/default/converge.yml
molecule/default/tasks/info.yml
plugins/doc_fragments/k8s_wait_options.py
plugins/modules/k8s.py
plugins/modules/k8s_info.py

### DIFF
--- a/molecule/default/tasks/info.yml
+++ b/molecule/default/tasks/info.yml
@@ -1,0 +1,164 @@
+---
+- block:
+    - set_fact:
+        wait_namespace: wait
+        k8s_pod_name: pod-info-1
+        multi_pod_one: multi-pod-1
+        multi_pod_two: multi-pod-2
+
+    - name: Ensure namespace exists
+      k8s:
+        definition:
+          apiVersion: v1
+          kind: Namespace
+          metadata:
+            name: "{{ wait_namespace }}"
+
+    - name: Add a simple pod with initContainer
+      k8s:
+        definition:
+          apiVersion: v1
+          kind: Pod
+          metadata:
+            name: "{{ k8s_pod_name }}"
+            namespace: "{{ wait_namespace }}"
+          spec:
+            initContainers:
+              - name: init-01
+                image: python:3.7-alpine
+                command: ['sh', '-c', 'sleep 20']
+            containers:
+              - name: utilitypod-01
+                image: python:3.7-alpine
+                command: ['sh', '-c', 'sleep 360']
+
+    - name: Wait and gather information about new pod
+      k8s_info:
+        name: "{{ k8s_pod_name }}"
+        kind: Pod
+        namespace: "{{ wait_namespace }}"
+        wait: yes
+        wait_sleep: 5
+        wait_timeout: 400
+      register: wait_info
+
+    - name: Assert that pod creation succeeded
+      assert:
+        that:
+          - wait_info is successful
+          - wait_info.changed is false
+
+    - name: Remove Pod
+      k8s:
+        api_version: v1
+        kind: Pod
+        name: "{{ k8s_pod_name }}"
+        namespace: "{{ wait_namespace }}"
+        state: absent
+        wait: yes
+      ignore_errors: yes
+      register: short_wait_remove_pod
+
+    - name: Check if pod is removed
+      assert:
+        that:
+          - short_wait_remove_pod is successful
+          - short_wait_remove_pod.changed is true
+
+    - name: Create multiple pod with initContainer
+      k8s:
+        definition:
+          apiVersion: v1
+          kind: Pod
+          metadata:
+            labels:
+              run: multi-box
+            name: "{{ multi_pod_one }}"
+            namespace: "{{ wait_namespace }}"
+          spec:
+            initContainers:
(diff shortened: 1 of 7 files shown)


plugins/module_utils/common.py
```diff
@@ -249,16 +264,34 @@ def find_resource(self, kind, api_version, fail=False):
...
+
         try:
             result = resource.get(name=name,
                                   namespace=namespace,
                                   label_selector=','.join(label_selectors),
-                                  field_selector=','.join(field_selectors)).to_dict()
-        except openshift.dynamic.exceptions.NotFoundError:
+                                  field_selector=','.join(field_selectors))
+            satisfied_by = []
+            if wait and result['kind'].endswith('List') and hasattr(result, 'items'):
```
fabianvf, line 284: you can do 
```
from openshift.dynamic import ResourceList
...
if isinstance(result, ResourceList):
...
```
which should avoid getting tripped up by an odd custom resource

plugins/module_utils/common.py
```diff
@@ -249,16 +264,34 @@ def find_resource(self, kind, api_version, fail=False):
...
+                # We have a list of ResourceInstance
+                for resource_instance in result.get('items'):
```
fabianvf, line 286: any reason not to just iterate on `result.items`?

plugins/module_utils/common.py
```diff
@@ -249,16 +264,34 @@ def find_resource(self, kind, api_version, fail=False):
...
+                    success, res, duration = self.wait(resource, resource_instance, sleep=wait_sleep, timeout=wait_timeout)
+                    if not success:
+                        self.fail(msg="Failed to gather information about %s(s)" % result.get('kind')[:-4])
```
fabianvf, line 289: `result.base_kind` should have the proper kind in it if you want to avoid the string manipulation

plugins/module_utils/common.py
fabianvf, line 284: I don't see the case for when `wait` is specified and the result is not a `ResourceList`

fabianvf (changes requested):

plugins/module_utils/common.py
Akasurde: I thought this before but `resource.get` returns `ResourceInstance` and not `ResourceList`.

plugins/module_utils/common.py
Akasurde: `result` is `ResourceInstance` and does not have `base_kind` (which is for ResourceList).

`res` is a dictionary since `self.wait` returns the dictionary.

plugins/module_utils/common.py
Akasurde: Done.

plugins/module_utils/common.py
Akasurde: ResourceInstance['items'] handles both single and multiple resources.

molecule/default/tasks/info.yml
```diff
@@ -0,0 +1,164 @@
...
+      k8s_info:
+        name: "{{ k8s_pod_name }}"
+        kind: Pod
+        namespace: "{{ wait_namespace }}"
+        wait: yes
+        wait_sleep: 5
+        wait_timeout: 400
+      register: wait_info
+
+    - name: Assert that pod creation succeeded
```
fabianvf, line 45: We should also verify that the pod is actually running

plugins/module_utils/common.py
```diff
@@ -249,22 +278,42 @@ def find_resource(self, kind, api_version, fail=False):
...
                                   label_selector=','.join(label_selectors),
-                                  field_selector=','.join(field_selectors)).to_dict()
-        except openshift.dynamic.exceptions.NotFoundError:
+                                  field_selector=','.join(field_selectors))
+            if wait:
+                satisfied_by = []
+                if isinstance(result, ResourceInstance):
+                    # We have a list of ResourceInstance
+                    for resource_instance in result.get('items', []):
+                        success, res, duration = self.wait(resource, resource_instance,
```
fabianvf, line 302: It still looks like we're only calling `self.wait` when we get a `ResourceList`, but not when we are getting just a single `ResourceInstance` back

fabianvf (changes requested):

molecule/default/tasks/info.yml
Akasurde: OK

plugins/module_utils/common.py
Akasurde: Done.

plugins/module_utils/common.py
fabianvf: Wow I'm an idiot, yeah you're totally right

plugins/module_utils/common.py
fabianvf: yep ignore all the comments about ResourceLists

fabianvf (approved):

PR #239, opened 2020-09-28
[WIP] Add k8s_expose module

fabianvf:
##### SUMMARY

Adds a new module named `k8s_expose`, which can create a v1.Service object for a variety of resources.

This module will also serve as the base for https://github.com/ansible-collections/community.okd/issues/31

Currently it is set up only to create Services, if we wanted to we could extend it to be able to create ingresses as was discussed here for the kubectl client: https://github.com/kubernetes/kubernetes/issues/43667

They decided against expose for the CLI, but it may make more sense for Ansible. This could be done either now or as a followup. The `okd` module would then just create routes instead of ingresses, but would otherwise be equivalent.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

k8s_expose

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/k8s_expose.py
+++ b/plugins/modules/k8s_expose.py
@@ -1,0 +1,224 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright (c) 2020, Red Hat
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+
+DOCUMENTATION = r'''
+module: k8s_expose
+
+short_description: Expose a resource as a new Kubernetes service. Roughly equivalent to kubectl expose.
+
+version_added: "1.1.0"
+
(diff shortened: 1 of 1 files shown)


geerlingguy:
@fabianvf - After the discussion over here (https://github.com/ansible-collections/community.okd/issues/31#issuecomment-700922955), what do you think about the idea of renaming to `k8s_ingress`?

fabianvf:
@geerlingguy this one is actually in line with `kubectl expose`, `k8s_ingress` would create `Ingress` resources rather than `Service`s and would be more in line with the proposed `kubectl create ingress` command

fabianvf:
This isn't really that useful, k8s_service can fill this need and k8s_ingress is the more interesting one

tima:
> This isn't really that useful, k8s_service can fill this need and k8s_ingress is the more interesting one

@fabianvf Can you clarify this comment. Are you suggesting we need to create a k8s_ingress modules here?

fabianvf:
@tima need is a strong word, but if we wanted to add a feature that would be useful `k8s_ingress` could provide a ton of value

PR #247, opened 2020-10-02
Add appVersion check to Helm

### LINKED ISSUE #246: Helm should include appVersion in it's idempotence check
##### SUMMARY

When re-deploying a Helm chart using Ansible I noticed no deployment was preformed and the console output listed my task as `ok` instead of `changed`. It appears that the Helm module only uses the `chart` key to determine if a Chart needs to be upgraded during execution, [see here](https://github.com/ansible-collections/community.kubernetes/blob/main/plugins/modules/helm.py#L498). This check should consider the `appVersion` specification of the chart as well.

I saw a few other related issues looking to implemented `helm-diff` as well which be a more complete solution.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

community.kubernetes.helm

##### ANSIBLE VERSION

```paste below
ansible 2.9.12
  config file = None
  configured module search path = ['/<path>/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /<path>/.venv/lib/python3.8/site-packages/ansible
  executable location = /<path>/.venv/bin/ansible
  python version = 3.8.2 (default, Jul 16 2020, 14:00:26) [GCC 9.3.0]
```

##### CONFIGURATION

```paste below
```

##### STEPS TO REPRODUCE

Modify a chart's `appVersion` and re-run the Helm task. Observe the Helm chart is not updated even though the Chart specification has changed.

##### EXPECTED RESULTS

Charts that change the `appVersion` key should be re-deployed.

##### ACTUAL RESULTS

The deployed release remains unchanged.

nat45928:
### SUMMARY

When upgrading a Helm release include the chart's appVersion in the idempotence check.

Fixes #246

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

helm

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/helm.py
+++ b/plugins/modules/helm.py
@@ -520,13 +520,21 @@ def main():
                               create_namespace=create_namespace, replace=replace)
             changed = True
 
-        elif force or release_values != release_status['values'] \
-                or (chart_info['name'] + '-' + chart_info['version']) != release_status["chart"] \
-                or chart_info['appVersion'] != release_status["app_version"]:
-            helm_cmd = deploy(helm_cmd, release_name, release_values, chart_ref, wait, wait_timeout,
-                              disable_hook, force, values_files=values_files, atomic=atomic,
-                              create_namespace=create_namespace, replace=replace)
-            changed = True
+        else:
+            # the 'appVersion' specification is optional in a chart
+            chart_app_version = chart_info.get('appVersion', None)
+            released_app_version = release_status.get('app_version', None)
+
+            # when deployed without an 'appVersion' chart value the 'helm list' command will return the entry `app_version: ""`
+            appversion_is_same = (chart_app_version == released_app_version) or (chart_app_version is None and released_app_version == "")
+
+            if force or release_values != release_status['values'] \
+                    or (chart_info['name'] + '-' + chart_info['version']) != release_status["chart"] \
+                    or not appversion_is_same:
+                helm_cmd = deploy(helm_cmd, release_name, release_values, chart_ref, wait, wait_timeout,
+                                  disable_hook, force, values_files=values_files, atomic=atomic,
+                                  create_namespace=create_namespace, replace=replace)
+                changed = True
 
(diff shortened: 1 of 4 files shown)


nat45928:
This is my first time contributing to anything Ansible related.

There does not appear to be any unit or integration tests for the `helm` module so please let me know if anything else is needed for this change.

nat45928:
@tima I've added some tests for this change. The `appVersion` of the chart can only be changed in the `Chart.yaml` file or on the command line using the `package` subcommand. I've added some tasks which package two chart archives with different appVersion values, then run the standard `tests_chart.yaml` tasks file.

Let me know if you think this is not sufficient or if anything else is needed for this change to be approved.

geerlingguy (approved):

plugins/modules/helm.py
```diff
@@ -496,7 +496,8 @@ def main():
             changed = True
 
         elif force or release_values != release_status['values'] \
-                or (chart_info['name'] + '-' + chart_info['version']) != release_status["chart"]:
+                or (chart_info['name'] + '-' + chart_info['version']) != release_status["chart"] \
+                or chart_info['appVersion'] != release_status["app_version"]:
```
fabianvf, line 500: Are these values guaranteed to be present?

geerlingguy, line 500: If an `appVersion` is not provided in the chart, would this fail gracefully?

nat45928: More discussion in https://github.com/ansible-collections/community.kubernetes/pull/247#discussion_r500382019

nat45928: Good call, the `appVersion` key is optional. I've added a second chart and associated tests.

nat45928:
There's one more issue with CI but I don't believe it is related to this change.

geerlingguy:
@nat45928 - Yeah that failure is occurring in our CI job as well.

PR #258, opened 2020-10-08
Embedded docs clarifications and refinements.

tima:
##### SUMMARY

This PR refines the embedded documentation in a number of plugins in this collection.

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

- plugins/inventory/k8s.py
- plugins/modules/k8s.py
- plugins/modules/k8s_cluster_info.py
- plugins/modules/k8s_rollback.py
- plugins/modules/helm_plugin
- plugins/modules/helm_plugin_info

### DIFF
--- a/plugins/inventory/k8s.py
+++ b/plugins/inventory/k8s.py
@@ -14,8 +14,9 @@ DOCUMENTATION = '''
     short_description: Kubernetes (K8s) inventory source
 
     description:
-      - Fetch containers and services for one or more clusters
-      - Groups by cluster name, namespace, namespace_services, namespace_pods, and labels
+      - Fetch containers and services for one or more clusters.
+      - Groups by cluster name, namespace, namespace_services, namespace_pods, and labels.
+      - Uses the kubectl connection plugin to access the Kubernetes cluster.
       - Uses k8s.(yml|yaml) YAML configuration file to set parameter values.
 
     options:
--- a/plugins/modules/k8s.py
+++ b/plugins/modules/k8s.py
@@ -179,17 +179,12 @@ EXAMPLES = r'''
     state: present
     definition: "{{ lookup('file', '/testing/deployment.yml') | from_yaml }}"
 
-- name: Read definition file from the Ansible controller file system after Jinja templating
-  community.kubernetes.k8s:
-    state: present
-    definition: "{{ lookup('template', '/testing/deployment.yml') | from_yaml }}"
-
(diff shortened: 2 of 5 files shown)


plugins/inventory/k8s.py
```diff
@@ -14,8 +14,9 @@
     short_description: Kubernetes (K8s) inventory source
 
     description:
-      - Fetch containers and services for one or more clusters
-      - Groups by cluster name, namespace, namespace_services, namespace_pods, and labels
+      - Fetch containers and services for one or more clusters.
+      - Groups by cluster name, namespace, namespace_services, namespace_pods, and labels.
+      - Uses the kubectl connection plugin to access the Kubernetes cluster.
```
geerlingguy, line 19: Is this line relevant for the end-user to know, or is it just an internal implementation detail?

plugins/modules/k8s.py
```diff
@@ -179,17 +179,12 @@
     state: present
     definition: "{{ lookup('file', '/testing/deployment.yml') | from_yaml }}"
 
-- name: Read definition file from the Ansible controller file system after Jinja templating
```
geerlingguy, line 182: Do we still want to document this as an alternative?

plugins/modules/k8s_rollback.py
```diff
@@ -17,7 +17,7 @@
 description:
   - Use the OpenShift Python client to perform the Rollback.
   - Authenticate using either a config file, certificates, password or token.
-  - Similar to the kubectl rollout undo command.
+  - Similar to the `kubectl rollout undo` command.
```
geerlingguy, line 20: Do backticks work here or do we need that fancy RST thing like `C()` (I don't know RST well enough).

plugins/inventory/k8s.py
tima: I thought it might be relevant to the end user. The connection plugin is the only plugin that requires `kubectl` rather than use the API directly. So I thought it was worth a mention in case the run into a problem related to this.

plugins/modules/k8s.py
tima: I wanted to encourage users to use the new feature. Also, the example above sort of shows that.

plugins/modules/k8s_rollback.py
tima: Oh right. Damn. I was thinking Markdown and not RST.

PR #259, opened 2020-10-09
Fix the downstream build and release process

### LINKED ISSUE #254: Prepare for a 1.1.0 release
##### SUMMARY

Today we'd like to release the 1.1.0 version of this collection—at the same time, we will release an identical 1.1.0 version as `kubernetes.core` in preparation for the moving of all of this collection's content into that namespace (see https://github.com/ansible-collections/community.kubernetes/issues/221 for details).

I will need to go through past PRs and make sure they are all accounted for in changelog fragments, then I'll coordinate with @tima on releasing it and making sure it's also working in the other namespace.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

Releases

##### ADDITIONAL INFORMATION

N/A

geerlingguy:
##### SUMMARY

Fixes #254.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

Makefile and utils/downstream.sh

##### ADDITIONAL INFORMATION

N/A

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 # Also needs to be updated in galaxy.yml
-VERSION = 1.0.0
+VERSION = 1.1.0
 
 TEST_ARGS ?= ""
 PYTHON_VERSION ?= `python -c 'import platform; print("{0}.{1}".format(platform.python_version_tuple()[0], platform.python_version_tuple()[1]))'`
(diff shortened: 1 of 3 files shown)


geerlingguy:
Hmm... so that `sed` change definitely doesn't work on Linux. Drat.

I found [this answer](https://stackoverflow.com/a/22084103/100134) which creates a bunch of `.bak` files and is cross-platform. I could add one more line after all the `sed` invocations finding and deleting all `.bak` files in the collection directory...

PR #270, opened 2020-10-12
PoC: ansible_module.turbo integration

goneri:
Aloha!

This PoC PR introduces the use of ansible_module.turbo[1] to speed up the
`k8s_info` module. It breaks everything else :-).

This two recording runs the `k8s_facts` 20 times in row. The first one
use the current `main` branch, the second use this PR.

- https://asciinema.org/a/364880: 21.2s
- https://asciinema.org/a/364881: 7.2s

This PR improves performance at two level.

Reuse the same client object
============================

With ansible_module.turbo, the module are actually run in a long running
Python process. Thanks to it, we can cache large Python objects. In this case,
we focus on the `openshift.dynamic.DynamicClient` instance.
We use the `lru_cache`[2] decorator to ensure we will always get the
same python object, as soon as we use the same authentication
parameters.
Everytime we call `get_api_client()` in a new task, we get the same
Python client that is already in memory.

Speed up the import
===================

The loading of the OpenShift SDK take a lot of time, and this is a
penality we pay everytime we start a new task:

```
$ time python3 -c "from openshift.dynamic import DynamicClient"

real    0m0.653s
user    0m0.966s
sys     0m0.469s
```

This PR try to defeer all the non-critical imports. This allow us to load
the modules once in the long running Python process. We can reuse the
same module task after task.

About this PR changes
=====================

The use of object complexify the ansible_module.turbo integration.
Currently, everytime we instanciate the `K8sAnsibleMixin` object we load all the SDK.
In addition, everytime we run a task, we end up with a new
`K8sAnsibleMixin` instance that can hardly cache.

These are the reasons why the flatten the nested structure.

If you like the approach, I will be happy to start a larger refactoring
to generalize the use of ansible_module.turbo.

- [1] https://github.com/ansible-collections/cloud.common/blob/main/README_ansible_turbo.module.rst
- [2] https://docs.python.org/3/library/functools.html#functools.lru_cache

### DIFF
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -8,7 +8,8 @@ authors:
   - willthames (https://github.com/willthames)
   - mmazur (https://github.com/mmazur)
   - jamescassell (https://github.com/jamescassell)
-dependencies: {}
+dependencies:
+  cloud.common: '*'
 description: Kubernetes Collection for Ansible.
 documentation: ''
 homepage: ''
--- a/plugins/module_utils/common.py
+++ b/plugins/module_utils/common.py
@@ -97,46 +97,6 @@ except ImportError as e:
     K8S_IMP_ERR = traceback.format_exc()
 
 
-def list_dict_str(value):
-    if isinstance(value, (list, dict, string_types)):
-        return value
-    raise TypeError
-
-
-ARG_ATTRIBUTES_BLACKLIST = ('property_path',)
-
-COMMON_ARG_SPEC = {
-    'state': {
-        'default': 'present',
-        'choices': ['present', 'absent'],
-    },
-    'force': {
-        'type': 'bool',
-        'default': False,
-    },
-}
-
-RESOURCE_ARG_SPEC = {
-    'resource_definition': {
-        'type': list_dict_str,
-        'aliases': ['definition', 'inline']
-    },
-    'src': {
-        'type': 'path',
-    },
-}
-
-NAME_ARG_SPEC = {
-    'kind': {},
-    'name': {},
-    'namespace': {},
-    'api_version': {
-        'default': 'v1',
-        'aliases': ['api', 'version'],
-    },
-}
-
-
 def get_api_client(module=None, **auth_params):
     auth_params = auth_params or {}
     auth = {}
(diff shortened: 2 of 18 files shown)


galaxy.yml
```diff
@@ -8,7 +8,8 @@ authors:
   - willthames (https://github.com/willthames)
   - mmazur (https://github.com/mmazur)
   - jamescassell (https://github.com/jamescassell)
-dependencies: {}
+dependencies:
+  cloud.common: '*'
```
geerlingguy, line 12: Interesting—it seems like maybe `ansible-test sanity` is not automatically installing this dependency?

goneri:
Another example, this time with `k8s_service` and `k8s_info`. I run each module 20 times.

- before https://asciinema.org/a/364924 38.8s
- after: https://asciinema.org/a/364926 12s

galaxy.yml
goneri: Indeed, the `ERROR: plugins/modules/k8s_info.py:145:0: traceback: ModuleNotFoundError: No module named 'ansible_collections.cloud'` error is suspicious.

plugins/module_utils/common.py
```diff
@@ -371,7 +298,6 @@ def fail(self, msg=None):
         self.fail_json(msg=msg)
 
     def _wait_for(self, resource, name, namespace, predicate, sleep, timeout, state):
-        start = datetime.now()
```
fabianvf, line 374: was this removed intentionally?

goneri: hum no indeed.

plugins/modules/k8s_info.py
```diff
@@ -142,58 +142,50 @@
...
-    K8sAnsibleMixin, AUTH_ARG_SPEC, WAIT_ARG_SPEC)
-
-
-class KubernetesInfoModule(K8sAnsibleMixin):
-
-    def __init__(self, *args, **kwargs):
-        module = AnsibleModule(
-            argument_spec=self.argspec,
-            supports_check_mode=True,
+from ansible_collections.cloud.common.plugins.module_utils.turbo.module import (
```
Akasurde, line 145: I would like to see a toggle to change behavior.

goneri: At this stage in the module loading, we cannot access the module parameters yet. As an alternative, I can use an environment variable or just fallback on AnsibleModule in case of ImportError.

goneri:
See: https://github.com/ansible-collections/community.kubernetes/pull/313

PR #291, opened 2020-11-04
test: older_openshift_fail.yml fails with py3.7+

goneri:
The `older_openshift_fail.yml` test playbook fails with Python 3.7
because the old `kubernetes.client.apis.admissionregistration_api`
module uses the new `async` keyword as a function parameter name.

From: https://docs.python.org/3/whatsnew/3.7.html

Backwards incompatible syntax changes:

    `async` and `await` are now reserved keywords.

In this case, we get a `SyntaxError` exception instead of the expected
`Failed to import the required Python library (openshift >= 0.7.2)`.

### DIFF
--- a/tests/integration/targets/kubernetes/tasks/main.yml
+++ b/tests/integration/targets/kubernetes/tasks/main.yml
@@ -73,20 +73,26 @@
   no_log: yes
 
 # Test graceful failure for older versions of openshift
+- name: Get the Python version
+  setup:
+    filter: ansible_python_version
 
-- pip:
-    name:
-      - openshift==0.6.0
-      - kubernetes==6.0.0
-      - coverage
-    virtualenv: "{{ virtualenv }}"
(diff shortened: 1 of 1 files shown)


goneri:
Can we get this one merged?

tima:
@geerlingguy and I were discussing this and think the test can just be nuked. It served its purpose, but  now it's checking for something rather old.

geerlingguy:
Yeah—it was 2018 when the last major release of 0.6.x came out. We're still testing for a bug (that I remember from then) that's been resolved for over two years of new releases...

I believe either @willthames or @fabianvf originally wrote that test though, and might have more missing context—it might be nice to get one final "yes" from one of them before totally nuking it though.

tima:
I'm going to merge this for now, but as mentioned earlier and in #300 we'd like to remove this test entirely.

PR #293, opened 2020-11-06
test: cp command instead of copy module because of selinux

goneri:
If the host uses SELinux, we need to have the system selinux module
in the virtualenv, otherwise the copy module will be broken.
We instead use the cp command to avoid that.

### DIFF
--- a/tests/integration/targets/kubernetes/tasks/validate_installed.yml
+++ b/tests/integration/targets/kubernetes/tasks/validate_installed.yml
@@ -5,9 +5,10 @@
         name: "{{ playbook_namespace }}"
         kind: Namespace
 
-    - copy:
-        src: files
-        dest: "{{ remote_tmp_dir }}"
+    # If the host uses SELinux, we need to have the system selinux module
+    # in the virtualenv, otherwise the copy module will be broken.
+    # We instead use the cp command to avoid that.
+    - command: cp -r files {{ remote_tmp_dir }}
 
     - name: incredibly simple ConfigMap
       k8s:


tests/integration/targets/kubernetes/tasks/validate_installed.yml
```diff
@@ -5,9 +5,10 @@
         name: "{{ playbook_namespace }}"
         kind: Namespace
 
-    - copy:
-        src: files
-        dest: "{{ remote_tmp_dir }}"
+    # If the host uses SELinux, we need to have the system selinux module
+    # in the virtualenv, otherwise the copy module will be broken.
+    # We instead use the cp command to avoid that.
+    - command: cp -r files {{ remote_tmp_dir }}
```
tima, line 11: Do we need to make that an optional dependency or something in the docs? Using `command` and `shell` for something we have a module for makes me cringe. It's like fingernails on a chalkboard really.

geerlingguy: I'd rather not incorporate this change for one environment-specific test dependency. Instead, we could just recommend that people who want to run the test suite in a virtualenv should install the system selinux module inside?

tima: @geerlingguy: That's is what I was thinking also.

goneri: The problem is that we build the virtualenvs during the test execution. So we cannot just create them manually before.
Another approach that I took initially was to build the virtualenv with site-packages to True when selinux was active, it works too.

PR #294, opened 2020-11-06
molecule: use the new charts repository

goneri:
molecule: use the new charts repository

Since October, 26th, the sable repository has a new location.
See: https://helm.sh/blog/new-location-stable-incubator-charts/

Without this change, the test-suite fails because of the following
error:

```
$ `helm repo add test_helm_repo https://kubernetes-charts.storage.googleapis.com`
Error: repo "https://kubernetes-charts.storage.googleapis.com" is no longer available; try "https://charts.helm.sh/stable" instead
```

### DIFF
--- a/molecule/default/roles/helm/defaults/main.yml
+++ b/molecule/default/roles/helm/defaults/main.yml
@@ -9,7 +9,7 @@ tiller_cluster_role: cluster-admin
 chart_test: "nginx-ingress"
 chart_test_version: 1.32.0
 chart_test_version_upgrade: 1.33.0
-chart_test_repo: "https://kubernetes-charts.storage.googleapis.com"
+chart_test_repo: "https://charts.helm.sh/stable"
 chart_test_git_repo: "http://github.com/helm/charts.git"
 chart_test_values:
   revisionHistoryLimit: 0
--- a/molecule/default/roles/helm/tasks/tests_chart/from_url.yml
+++ b/molecule/default/roles/helm/tasks/tests_chart/from_url.yml
@@ -3,5 +3,7 @@
   include_tasks: "../tests_chart.yml"
   vars:
     source: url
-    chart_source: "{{ chart_test_repo }}/{{ chart_test }}-{{ chart_test_version }}.tgz"
-    chart_source_upgrade: "{{ chart_test_repo }}/{{ chart_test }}-{{ chart_test_version_upgrade }}.tgz"
+    chart_test: "fission"
+    cjart_test_version: "0.10.0"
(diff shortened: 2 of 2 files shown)


geerlingguy:
@goneri - Strange... if I load https://kubernetes-charts.storage.googleapis.com I can get the XML just fine.

Also, if I look at the https://charts.helm.sh/stable/ page, it mentions a [Deprecation and Archive Notice](https://github.com/helm/charts/blob/master/README.md#%EF%B8%8F-deprecation-and-archive-notice), which seems to indicate it shouldn't be used for anything moving forward.

Maybe we should pick some other repo that has a better chance of sticking around a long time?

goneri:
> @goneri - Strange... if I load https://kubernetes-charts.storage.googleapis.com I can get the XML just fine.
> 
> Also, if I look at the https://charts.helm.sh/stable/ page, it mentions a [Deprecation and Archive Notice](https://github.com/helm/charts/blob/master/README.md#%EF%B8%8F-deprecation-and-archive-notice), which seems to indicate it shouldn't be used for anything moving forward.
> 
> Maybe we should pick some other repo that has a better chance of sticking around a long time?

Hum weird. I hit another wall after that, nginx-ingress-1.32.0.tgz don't seem to be online anymore. But well, It may totally be because of my set-up. It's the first time I try to run the test-suite.

goneri:
```
[goneri@t580-gone kubernetes]$ helm repo add test_helm_repo https://kubernetes-charts.storage.googleapis.com
Error: repo "https://kubernetes-charts.storage.googleapis.com" is no longer available; try "https://charts.helm.sh/stable" instead
```

goneri:
The new location has been announced here: https://helm.sh/blog/new-location-stable-incubator-charts/

goneri:
There is still some problem, but are you ok with this strategy?

geerlingguy:
I think if we're going to move to a different charts repository, we should just move to one that we know will be maintained/stable, and not move to https://charts.helm.sh/stable/, which says it is deprecated already and on November 13 (less than a week away) it will stop working entirely...

goneri:
> I think if we're going to move to a different charts repository, we should just move to one that we know will be maintained/stable, and not move to https://charts.helm.sh/stable/, which says it is deprecated already and on November 13 (less than a week away) it will stop working entirely...

Oh, good point. The new home of ingress-nginx seems to be https://kubernetes.github.io/ingress-nginx. I give it a try.

geerlingguy:
Thanks for all the work on this—closing in favor of #297

PR #304, opened 2020-11-12
test: reduce the cluster's min number of pods

goneri:
With minishift, there is just 3 pods running.

### DIFF
--- a/molecule/default/converge.yml
+++ b/molecule/default/converge.yml
@@ -16,9 +16,9 @@
         kind: Pod
       register: pod_list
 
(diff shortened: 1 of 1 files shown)


goneri:
Please don't merge. I'm actually not sure if this is a normal situation.

geerlingguy:
@goneri - Can you try switching this PR to 'Draft' just to make it extra certain someone doesn't merge accidentally?

goneri:
Closing, I don't need that with Kind BTW.

PR #310, opened 2020-11-24
The type of wait_condition.status is str

goneri:
The `wait_condition.status` key is a string.

- Use actually string in the documentation
- Use core's `boolean()` method to convert the value internally to
  boolean

### DIFF
--- a/plugins/module_utils/common.py
+++ b/plugins/module_utils/common.py
@@ -30,6 +30,8 @@ from ansible.module_utils.basic import AnsibleModule, missing_required_lib
 from ansible.module_utils.six import iteritems, string_types
 from ansible.module_utils._text import to_native
 from ansible.module_utils.common.dict_transformations import dict_merge
+from ansible.module_utils.parsing.convert_bool import boolean
+
 
 K8S_IMP_ERR = None
 try:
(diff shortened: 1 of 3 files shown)


plugins/module_utils/common.py
```diff
@@ -437,8 +439,8 @@ def _custom_condition(resource):
                     if condition['reason']:
                         return match.reason == condition['reason']
                 return False
-            status = True if match.status == 'True' else False
-            if status == condition['status']:
+            if status = True if match.status == 'True' else False
```
geerlingguy, line 442: Maybe this is my python not being amazing here, but this line seems wrong to me...?

goneri: Damn, not it's me messing my `git rebase` :-)

geerlingguy (approved):

plugins/module_utils/common.py
geerlingguy: Okay good... I was like "I'm no Python God, and I know there can be some weird inline stuff (coming from PHP I've seen it all)... but that just doesn't seem like it would do anything good."

goneri: Well, I did 10 years of Perl-fu in a previous life :D.

changelogs/fragments/wait_condition.status_is_a_str.yaml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
```
geerlingguy, line 2: Can you rename the fragment to `310-wait_condition.status_is_a_str.yaml` (just to keep similar format to all the others with issue/PR in the name).

goneri: Done.

PR #313, opened 2020-11-30
refactoring for ansible_module.turbo integration

goneri:
This refactoring prepares the integration of `ansible_module.turbo`

- Delay the loading of `common.py`, move the shared structure in
  `args_common`.
- Avoid the use of one single object per module, this to increase the
  amount of Python structure that we can cache.

See: https://github.com/ansible-collections/community.kubernetes/pull/270


plugins/module_utils/common.py
```diff
@@ -98,112 +99,83 @@
...
+    auth = {}
+
+    import hashlib
+    m = hashlib.sha256()
+    for i in AUTH_ARG_MAP:
+        m.update(str(auth_params.get(i)).encode())
+    digest = m.hexdigest()
+    if digest in get_api_client._pool:
+        client = get_api_client._pool[digest]
+        return client
```
fabianvf, line 113: I don't think we can correctly return here, as it's possible that the content of the kubeconfig, the content of the incluster config, or the envvars could have changed since the last time this was called. I think we need to wait until we've pulled the variables from all possible locations to create this digest. IOW, the kubernetes.client.Configuration object would ideally be the object that is used to create this digest.

plugins/module_utils/common.py
```diff
@@ -98,112 +99,83 @@
...
+    except AttributeError:
+        configuration = kubernetes.client.Configuration()
+
+    for key, value in iteritems(auth):
+        if key in AUTH_ARG_MAP.keys() and value is not None:
+            if key == 'api_key':
+                setattr(configuration, key, {'authorization': "Bearer {0}".format(value)})
+            else:
+                setattr(configuration, key, value)
+
```
fabianvf, line 167: This is the point at which we have a complete configuration object (`configuration`)

plugins/module_utils/common.py
```diff
@@ -98,112 +99,83 @@
...
+    try:
+        return DynamicClient(kubernetes.client.ApiClient(configuration))
```
fabianvf, line 169: I think we don't want to return here and should instead put line 173 inside this block

fabianvf (changes requested):
changes overall look good, just saw an issue with when the api client is being cached

plugins/module_utils/common.py
goneri: thank you @fabianvf for the review. I agree, it's a much better strategy. I just pushed a new revision of the patch that uses this strategy.

fabianvf (dismissed):

gravesm:
Per @fabianvf we should make sure that these changes don't impact the okd collection, as it imports a fair amount from this one.

fabianvf:
@gravesm I think the impact is almost certain, we'd just need to ensure we've got pins in place on the okd collection to not pull in an updated community.kubernetes collection until it has been updated. Don't think it needs to block this PR though as long as the work is on our radar and done before a community.kubernetes release would be cut.

gravesm:
@goneri can you rebase this when you get a chance? We should be able to get this reviewed and merged now that 1.2 has been released.

jillr:
I can't actually run sanity on this locally (I need to figure out what's up with that on my machine) but I've got a PR to @goneri's branch that I think should solve the failures here.

plugins/module_utils/ansiblemodule.py
```diff
@@ -0,0 +1,6 @@
+from __future__ import (absolute_import, division, print_function)
+
+__metaclass__ = type
+
+
+from ansible.module_utils.basic import AnsibleModule  # noqa: F401
```
gravesm, line 6: This is not blocking, but I'm curious why we are doing this?

plugins/module_utils/common.py
```diff
@@ -99,201 +100,107 @@
...
-    'cert_file': 'client_cert',
-    'key_file': 'client_key',
-    'proxy': 'proxy',
-    'persist_config': 'persist_config',
-}
-
+def configuration_digest(configuration):
+    import hashlib
+    m = hashlib.sha256()
+    for k in AUTH_ARG_MAP:
```
gravesm, line 106: I'm still unclear on which python versions we are supporting. If the intent is for this to work on 2.7 then I think we should use an OrderedDict to ensure the caching works. If not, feel free to ignore.

gravesm (commented):
I know there was some mention of this only working on python 3. If that's the case, what's the thing keeping it from working on python 2?

plugins/module_utils/ansiblemodule.py
goneri: This is the place where we will load the Turbo mode. I will use a conditional block to only enable it if we're running Py3.6+.

plugins/module_utils/common.py
goneri: Since we will only cache the session on py3.6+, so we can keep the code as it is today.

gravesm (approved):

PR #320, opened 2020-12-03
k8s: handle remote src and kubeconfig

Akasurde:
##### SUMMARY

Add logic to specify if
src file is located on a remote node or Ansible Controller.

Add logic to specify if 
kubeconfig file is located on the remote node or Ansible Controller.

Fixes: #307

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
changelogs/fragments/307_remote_src.yml
plugins/action/k8s_info.py

### DIFF
--- a/plugins/action/k8s_info.py
+++ b/plugins/action/k8s_info.py
@@ -42,12 +42,18 @@ class ActionModule(ActionBase):
         result = super(ActionModule, self).run(tmp, task_vars)
         del tmp  # tmp no longer has any effect
 
+        # Check current transport connection and depending upon
+        # look for kubeconfig and src
+        # 'local' => look files on Ansible Controller
+        # Transport other than 'local' => look files on remote node
+        remote_transport = self._connection.transport != 'local'
+
         new_module_args = copy.deepcopy(self._task.args)
 
         kubeconfig = self._task.args.get('kubeconfig', None)
-        remote_kubeconfig = self._task.args.get('remote_kubeconfig', False)
         # find the kubeconfig in the expected search path
-        if kubeconfig and not remote_kubeconfig:
+        if kubeconfig and not remote_transport:
+            # kubeconfig is local
             try:
                 # find in expected paths
                 kubeconfig = self._find_needle('files', kubeconfig)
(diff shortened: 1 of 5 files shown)


plugins/action/k8s_info.py
```diff
@@ -62,7 +62,14 @@ def run(self, tmp=None, task_vars=None):
 
         # find the file in the expected search path
         src = self._task.args.get('src', None)
+        remote_src = self._task.args.get('remote_src', False)
+
         if src:
+            if remote_src:
```
fabianvf, line 68: Do we want to do something for `kubeconfig` as well? It's a bit awkward but that is another file that we look for locally but could be present remotely.

Akasurde:
recheck

sylvainOL:
one of two issues of #307 is solved thanks to this PR. 👍

tima (changes requested):
I'm not a fan of this implementation. I'm pretty sure when I wrote the synchronize I was able to figure this out in the code without an extra param.

gravesm (approved):
LGTM

plugins/action/k8s_info.py
```diff
@@ -42,10 +42,15 @@ def run(self, tmp=None, task_vars=None):
         result = super(ActionModule, self).run(tmp, task_vars)
         del tmp  # tmp no longer has any effect
 
+        remote_transport = False
+        if self._connection.transport != 'local':
+            remote_transport = True
```
goneri, line 47: `remote_transport = self._connection.transport != 'local'`

plugins/action/k8s_info.py
```diff
@@ -42,10 +42,15 @@ def run(self, tmp=None, task_vars=None):
         result = super(ActionModule, self).run(tmp, task_vars)
         del tmp  # tmp no longer has any effect
 
+        remote_transport = False
+        if self._connection.transport != 'local':
```
fabianvf, line 46: Could you add some docs to the module describing exactly what the behavior users will see is? I don't know much about how Ansible handles all this stuff internally so I'm not really sure when it'll try to load it locally vs remotely

fabianvf (changes requested):

fabianvf (approved):

plugins/action/k8s_info.py
fabianvf: Changing my review to approving because I might not get back to this in a reasonable amount of time, but still requesting some docs

tima (approved):
LGTM once @goneri and @fabianvf comments are addressed.

goneri (approved):

goneri:
Thank you @Akasurde :-).

PR #324, opened 2020-12-07
helm and helm_info: adjust the name of two parameters

goneri:
According to the documentation fragment and the other modules:
- the `kube_context` option should be named `context`
- the `kubeconfig_path` option should be named `kubeconfig`.

PR #325, opened 2020-12-07
add 3 doc-default-does-not-match-spec errors

goneri:
Address the following error with the test-suite:

- plugins/modules/k8s.py:0:0: doc-default-does-not-match-spec: Argument 'append_hash' in argument_spec defines default as (False) but documentation defines default as (None)
- plugins/modules/k8s.py:0:0: doc-default-does-not-match-spec: Argument 'apply' in argument_spec defines default as (False) but documentation defines default as (None)
- plugins/modules/k8s_service.py:0:0: doc-default-does-not-match-spec: Argument 'apply' in argument_spec defines default as (False) but documentation defines default as (None)

### DIFF
--- a/plugins/modules/k8s.py
+++ b/plugins/modules/k8s.py
@@ -87,6 +87,7 @@ options:
       will only work if the same object is passed with state=absent (alternatively, just use state=absent with the name including
       the generated hash and append_hash=no)
     - Requires openshift >= 0.7.2
+    default: False
     type: bool
   apply:
     description:
@@ -95,6 +96,7 @@ options:
     - C(apply) works better with Services than 'force=yes'
     - Requires openshift >= 0.9.2
     - mutually exclusive with C(merge_type)
+    default: False
     type: bool
   template:
     description:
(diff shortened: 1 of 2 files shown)


geerlingguy:
@goneri - I believe @Akasurde was also working on this separately in one of his PRs... but I would like to have it separated anyways, so we can merge it independent of whatever other work is going on in PRs currently.

goneri:
> @goneri - I believe @Akasurde was also working on this separately in one of his PRs... but I would like to have it separated anyways, so we can merge it independent of whatever other work is going on in PRs currently.

Indeed, I just saw his fixes: https://github.com/ansible-collections/community.kubernetes/pull/320/files#diff-0f6f43c36182e7fd2b2d7f77c3dafc90ae60a782ed8831830916504505211b2b

I also prefer when a patch address a single issue.

plugins/modules/k8s.py
```diff
@@ -87,6 +87,7 @@
       will only work if the same object is passed with state=absent (alternatively, just use state=absent with the name including
       the generated hash and append_hash=no)
     - Requires openshift >= 0.7.2
+    default: False
```
tima, line 90: I'm a bit confused too because I just saw these same changes in the PR #324.

goneri: Yeah I didn't know about @Akasurde fix until @geerlingguy told me.

Akasurde (approved):
LGTM

PR #330, opened 2020-12-10
k8s: Handle ValueError raised

Akasurde:
##### SUMMARY

API raises ValueError when namespace is not provided in
definition. Handle and provide meaningful message to user
about this.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
changelogs/fragments/handle_valueerror.yml

### DIFF
--- a/plugins/module_utils/common.py
+++ b/plugins/module_utils/common.py
@@ -633,8 +633,8 @@ class K8sAnsibleMixin(object):
         except DynamicApiError as exc:
             self.fail_json(msg='Failed to retrieve requested object: {0}'.format(exc.body),
                            error=exc.status, status=exc.status, reason=exc.reason)
-        except Exception as exc:
-            self.fail_json(msg='Failed to retrieve requested object: {0}'.format(to_native(exc)),
+        except ValueError as value_exc:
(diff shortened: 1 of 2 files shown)


plugins/module_utils/common.py
```diff
@@ -708,6 +708,11 @@ def perform_action(self, resource, definition):
                         if self.warnings:
                             msg += "\n" + "\n    ".join(self.warnings)
                         self.fail_json(msg=msg, error=exc.status, status=exc.status, reason=exc.reason)
+                    except Exception as exc:
```
goneri, line 711: I would prefer we just catch the `ValueError`. This way if another valide exception is raised, it won't be hidden.

plugins/module_utils/common.py
```diff
@@ -708,6 +708,11 @@ def perform_action(self, resource, definition):
...
+                        msg = "Failed to create object: {0}".format(exc)
```
goneri, line 712: If we only catch the `ValueError`, we can show up a better message to the user. e.g: Missing or invalid parameter.

gravesm (approved):
lgtm

Issue #274, opened 2020-10-19
community.kubernetes.helm: release_values makes ansible always show changed state


SpikePy:
##### SUMMARY
When using `values_files: /tmp/{{ role_name }}/values_rabbitmq.yml` in the new helm module ansible always shows a _changed_ state after running it (even though the file has the same content), but if I use 
`release_values: "{{ lookup('template', 'values_rabbitmq.yml') | from_yaml }}"` with the same content it just shows ok and only shows change when something actually changed in the given values.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
community.kubernetes.helm
##### ANSIBLE VERSION

```paste below
ansible 2.10.2
  config file = None
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.9/site-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 3.9.0 (default, Oct 13 2020, 20:25:37) [GCC 8.3.0]
```

##### CONFIGURATION

```paste below
DEFAULT_LOAD_CALLBACK_PLUGINS(env: ANSIBLE_LOAD_CALLBACK_PLUGINS) = True
DEFAULT_STDOUT_CALLBACK(env: ANSIBLE_STDOUT_CALLBACK) = yaml
```

##### OS / ENVIRONMENT

Linux work 5.8.0-3-amd64 #1 SMP Debian 5.8.14-1 (2020-10-10) x86_64 GNU/Linux

##### STEPS TO REPRODUCE

```yaml
# rabbitmq
- name: Create temporary directory for values
  file:
    path: /tmp/{{ role_name }}
    state: directory
  changed_when: false

- name: Write configs to temporary file
  template:
    src: "{{ item }}"
    dest: /tmp/{{ role_name }}/{{ item | basename | regex_replace('\.j2$','') }}
  with_fileglob:
    - ../templates/*.j2
  changed_when: false

- name: Install rabbitmq cluster
  community.kubernetes.helm:
    chart_repo_url: https://charts.bitnami.com/bitnami
    chart_ref: rabbitmq
    release_namespace: "{{ rabbitmq_namespace }}"
    name: "{{ rabbitmq_install_name }}"
    chart_version: "{{ rabbitmq_helm_version }}"
    values_files: /tmp/{{ role_name }}/values_rabbitmq.yml
[...]

PR #332, opened 2020-12-10
helm: 'changed' flag takes 'values' in consideration

goneri:
Return `changed` `False` is the `values_files` match the `values` of
the existing deployment.

Closes: #274

### DIFF
--- a/plugins/modules/helm.py
+++ b/plugins/modules/helm.py
@@ -387,6 +387,18 @@ def delete(command, release_name, purge, disable_hook):
     return delete_command
 
 
+def load_values_files(values_files):
+    values = {}
+    for values_file in values_files or []:
+        with open(values_file, 'r') as fd:
+            content = yaml.safe_load(fd)
+            if not isinstance(content, dict):
+                continue
+            for k, v in content.items():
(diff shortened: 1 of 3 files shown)


tima (approved):
lgtm

gravesm (approved):
lgtm

plugins/modules/helm.py
```diff
@@ -502,7 +514,12 @@ def main():
             # when deployed without an 'appVersion' chart value the 'helm list' command will return the entry `app_version: ""`
             appversion_is_same = (chart_app_version == released_app_version) or (chart_app_version is None and released_app_version == "")
 
-            if force or release_values != release_status['values'] \
+            if values_files:
+                values_are_good = release_status['values'] == load_values_files(values_files)
```
geerlingguy, line 518: Only nit with this patch is this variable name—it's not like if they don't match they're _bad_. I would rather this variable be named `values_match` or something along those lines.

goneri: indeed. Good point.

geerlingguy (approved):

PR #334, opened 2020-12-17
Add support for configuring garbage collection

### LINKED ISSUE #253: Add a new `delete_options` parameter to the `k8s` module
### SUMMARY

The `k8s` module needs a way to implement a way to pass `deleteOptions` values through to the K8s API client. This feature request would address the problem reported in #215, where cascading deletes are needed, but cannot be performed with the modules today.

This would entail adding a new `delete_options` parameter to the `k8s` module that can pass through [a complex data structure with a schema like the `deleteOptions` in Kubernetes](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#deleteoptions-v1-meta).

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

plugins/modules/k8s.py

gravesm:
##### SUMMARY
This surfaces [deleteOptions](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#deleteoptions-v1-meta) functionality in a top-level `delete_options`
parameter. `delete_options` can be used to control garbage collection behavior at the time of resource deletion.

Closes #253.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
k8s.py

##### ADDITIONAL INFORMATION
A few things to note:
1. I have left off the `dryRun` parameter as it seems like this would conflict with ansible's check mode.
2. `apiVersion` and `kind` are included, but I'm not sure how useful they are in this context. `kind` in particular can only be `DeleteOptions`.

### DIFF
--- a/molecule/default/tasks/gc.yml
+++ b/molecule/default/tasks/gc.yml
@@ -1,0 +1,213 @@
+---
+- vars:
+    gc_namespace: garbage
+    gc_name: garbage-job
+    # This is a job definition that runs for 10 minutes and won't gracefully
+    # shutdown. It allows us to test foreground vs background deletion.
+    job_definition:
+      apiVersion: v1
+      kind: Job
+      metadata:
+        name: "{{ gc_name }}"
+        namespace: "{{ gc_namespace }}"
+      spec:
+        template:
+          metadata:
+            labels:
+              job: gc
+          spec:
+            containers:
+              - name: "{{ gc_name }}"
+                image: busybox
+                command:
+                  - sleep
+                  - "600"
+            restartPolicy: Never
+
+  block:
+    - name: Ensure namespace exists
+      k8s:
+        definition:
+          apiVersion: v1
+          kind: Namespace
+          metadata:
+            name: "{{ gc_namespace }}"
+
+    - name: Add a job
+      k8s:
+        definition: "{{ job_definition }}"
+
+    - name: Test that job's pod is running
+      k8s_info:
+        kind: Pod
+        namespace: "{{ gc_namespace }}"
+        label_selectors:
+          - "job=gc"
+        wait: yes
+        wait_timeout: 100
+      register: job
+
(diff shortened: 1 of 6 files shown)


molecule/default/tasks/gc.yml
```diff
@@ -0,0 +1,213 @@
...
+    - name: Delete job in foreground
+      k8s:
+        kind: Job
+        name: "{{ gc_name }}"
+        namespace: "{{ gc_namespace }}"
+        state: absent
+        wait: yes
+        wait_timeout: 100
+        delete_options:
+          apiVersion: v1
```
fabianvf, line 63: yeah I'd say drop apiVersion and kind here, that's more an API detail than something that we need to expose

tima: I agree with @fabianvf. We try to remove users from having to know and specify these things whenever we can.

tima (approved):
Other than the one comment by @fabianvf about `apiVersion` LGTM FWIW.

tima (approved):

plugins/doc_fragments/k8s_delete_options.py
```diff
@@ -0,0 +1,50 @@
...
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+
+class ModuleDocFragment(object):
+
+    DOCUMENTATION = r'''
+options:
+  delete_options:
+    type: dict
```
Akasurde, line 17:
```suggestion
    type: dict
    version_added: '1.2.0'
```

Akasurde: @tima @geerlingguy Can you confirm this?

tima: @Akasurde You mean the version of out next release? If so, 1.2.0 is the plan last we discussed.

Akasurde (approved):

PR #350, opened 2021-01-22
k8s: Support for directory in src

Akasurde:
##### SUMMARY

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME

### DIFF
--- a/plugins/module_utils/common.py
+++ b/plugins/module_utils/common.py
@@ -19,10 +19,10 @@ from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
 import base64
-import time
 import os
-import traceback
 import sys
+import time
+import traceback
 from datetime import datetime
 from distutils.version import LooseVersion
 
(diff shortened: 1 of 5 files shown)


plugins/module_utils/common.py
```diff
@@ -359,18 +361,53 @@ def remove_aliases(self):
...
 
+    def is_directory(self, path):
+        path = self.path_dwim(path)
+        return os.path.isdir(to_bytes(path, errors='surrogate_or_strict'))
+
+    def list_directory(self, path):
+        path = self.path_dwim(path)
+        return os.listdir(path)
+
+    def path_dwim(self, given):
```
gravesm, line 372: I think this function duplicates what's already in `unfrackpath`.

gravesm (commented):
It would probably be a good idea to add some tests, and it needs a changelog fragment. Otherwise, lgtm.

PR #355, opened 2021-01-27
Add optional support for helm diff

### LINKED ISSUE #248: Improve helm module idempotency checks with optional helm-diff support
##### SUMMARY

A number of issues have been filed where Ansible does not properly detect changes in a Helm chart depending on the nature of the changes. While improved logic could and should be added to the module, for the best and most complete results, helm-diff should be integrated into the modules logic. 

At this time, I am recommending this support should be made optional to avoid another hard dependency. The module should detect if helm-diff is available and use that tool if any change is present. If helm-diff is not present, the built into python checks are performed along with a warning message. 

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

plugins/modules/helm.py

##### ADDITIONAL INFORMATION

This feature requests comes out of [a comment in #136](https://github.com/ansible-collections/community.kubernetes/pull/136#discussion_r452350040) and addresses [the last remaining issue mentioned in #236](https://github.com/ansible-collections/community.kubernetes/pull/236#pullrequestreview-498624519). 

This work subsumes issues #128 and #207.

gravesm:
##### SUMMARY

There are some cases where the existing module has difficulty
determining if an upgrade would result in changes. This can particularly
be a problem when changes are made to a local chart.

This adds optional support for helm diff. If the plugin is present it
will be used. Otherwise, the default implementation will be used and a
warning will be issued. One caveat: helm diff does not currently support
using a repo url, so the default implementation will be used in this
case, as well.

Closes #248 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

helm

##### ADDITIONAL INFORMATION

Several of our existing tests break when helm diff is installed. This is because our local path test charts don't actually contain any resources so nothing changes when we do any upgrade. Changing the `appVersion` or adding new values won't make a difference. I have added a simple `ConfigMap` to our test charts so that when an upgrade happens it will actually change something. This is the reason for all the changes to the existing tests.

### DIFF
--- a/changelogs/fragments/355-helm-diff.yaml
+++ b/changelogs/fragments/355-helm-diff.yaml
@@ -1,0 +1,2 @@
+minor_changes:
+  - helm - add optional support for helm diff (https://github.com/ansible-collections/community.kubernetes/issues/248).
--- a/molecule/default/roles/helm/defaults/main.yml
+++ b/molecule/default/roles/helm/defaults/main.yml
@@ -16,3 +16,4 @@ chart_test_repo: "https://kubernetes.github.io/ingress-nginx"
 chart_test_git_repo: "http://github.com/helm/charts.git"
 chart_test_values:
   revisionHistoryLimit: 0
+  myValue: "changed"
--- a/molecule/default/roles/helm/files/appversionless-chart-v2/Chart.yaml
+++ b/molecule/default/roles/helm/files/appversionless-chart-v2/Chart.yaml
@@ -1,0 +1,5 @@
+apiVersion: v2
+name: appversionless-chart
+description: A chart used in molecule tests
+type: application
+version: 0.2.0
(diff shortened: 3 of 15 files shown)


Akasurde:
```
ERROR: plugins/modules/helm.py:412:14: blacklisted-name: Black listed name "_"
```

gravesm:
@Akasurde yes, that one I'm not worried about. It's an easy fix. The bigger problem is I don't know how to address the yamllint errors. Fixing those errors will break the templates for go.

Akasurde:
Oh yes. I didn't see that. I will check that now.

gravesm:
@Akasurde I think I sorted out the linting issues. If there's a better way to do this, I'm happy to be educated. I wish we could ignore a directory, but that doesn't seem possible.

Akasurde:
@gravesm We can disable yamllint with several ways.

1. Disable directory - https://yamllint.readthedocs.io/en/stable/configuration.html#ignoring-paths
2. Disable single line - https://yamllint.readthedocs.io/en/stable/disable_with_comments.html#disabling-checks-for-a-specific-line
3. Disable single file - https://yamllint.readthedocs.io/en/stable/disable_with_comments.html#disabling-checks-for-all-or-part-of-the-file

I am OK with current configuration as well since I don't see chart files changing any time soon.

Akasurde (approved):
LGTM

tima (approved):

Issue #357, opened 2021-01-28
helm_plugin requires release_namespace


ecksun:
##### SUMMARY
release_namespace is required in the helm_plugin task.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
community.kubernetes.helm_plugin

##### ANSIBLE VERSION

```paste below
ansible 2.10.3
  config file = /home/user/code/noomi/devops/ansible/cloud-infrastructure/ansible.cfg
  configured module search path = ['/home/user/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/user/code/noomi/devops/ansible/.venv/lib/python3.7/site-packages/ansible
  executable location = /home/user/code/noomi/devops/ansible/.venv/bin/ansible
  python version = 3.7.3 (default, Jul 25 2020, 13:03:44) [GCC 8.3.0]
```

##### CONFIGURATION

```paste below
INTERPRETER_PYTHON(/home/user/code/noomi/devops/ansible/cloud-infrastructure/ansible.cfg) = ../.venv/bin/python3
```

##### STEPS TO REPRODUCE

```yaml
- name: Install local Helm plugin for Google Cloud Storage repository support
  community.kubernetes.helm_plugin:
    plugin_path: https://github.com/hayorov/helm-gcs
    state: present
```

Looking at [the souce](https://github.com/Akasurde/community.kubernetes/blob/main/plugins/modules/helm_plugin.py#L105) this seems to affect even master.

I'm assuming that in many cases it make sense to have the argument required (perhaps it was more relevant with helm2?), but I don't think it does in this case.

##### EXPECTED RESULTS

The task should run without having to specify namespace, as the plugin is only local and doesn't deal with any namespaces.

##### ACTUAL RESULTS

```paste below
TASK [therole : Install local Helm plugin for Google Cloud Storage repository support] ********************************************************************************************************************************************************************************************************************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "missing required arguments: release_namespace"}
```

Akasurde:
@ecksun Thanks for reporting this. Do you think we should make this as an optional parameter?

ecksun:
> @ecksun Thanks for reporting this. Do you think we should make this as an optional parameter?

probably?

I'm not super familiar with helm and helm plugins so there might be something I haven't understood correctly or that I have missed. But consider my current task:

```yaml
[...]

PR #358, opened 2021-02-02
Make unused release_namespace parameter as optional

Akasurde:
##### SUMMARY

Modules modified -
* helm_plugin
* helm_plugin_info

Fixes: #357

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
changelogs/fragments/357_helm_plugin.yml
plugins/modules/helm_plugin.py


gravesm:
I'm fine with the change, but I wonder if it might be better to push this to 2.0 since it will break any existing playbooks using these modules? Maybe we should remove the `required` attribute of this parameter for 1.2, which would fix the reported bug, and then merge this change in 2.0?

Akasurde:
Make sense to me.

tima:
I agree with what @gravesm [said](https://github.com/ansible-collections/community.kubernetes/pull/358#issuecomment-771625071).

tima (approved):
I agree with making this optional in 1.2 and slating it for removal in 2.0.

PR #360, opened 2021-02-04
Honor wait_timeout in k8s_info for missing resource

### LINKED ISSUE #344: k8s_info with wait option returns immediately when resource doesn't exist
##### SUMMARY

`k8s_info` with the `wait` option does return immediately when no object is found. 

From the example in the

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

`community.kubernetes.k8s_info` (in collection v1.1.1)

##### ANSIBLE VERSION

```paste below
ansible 2.9.13
  config file = /Users/hojerst/.ansible.cfg
  configured module search path = ['/Users/hojerst/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /Users/hojerst/devel/.direnv/python-3.9.0/lib/python3.9/site-packages/ansible
  executable location = /Users/hojerst/devel/.direnv/python-3.9.0/bin/ansible
  python version = 3.9.0 (default, Nov 28 2020, 00:47:30) [Clang 12.0.0 (clang-1200.0.32.27)]
```

##### CONFIGURATION

```paste below
(none)
```

##### STEPS TO REPRODUCE

run the example from the documentation without a object "pod/pod-not-yet-created"

```yaml
- name: Wait till the Object is created
  community.kubernetes.k8s_info:
    kind: Pod
    wait: yes
    name: pod-not-yet-created
    namespace: default
    wait_sleep: 10
    wait_timeout: 360
```

##### EXPECTED RESULTS

I'm expecting that the module waits for up to 360 seconds until the resource "pod/pod-not-yet-created" is in the desired state (Running for Pods).

##### ACTUAL RESULTS

The module returns immediately without any waiting if the resource is not yet created. This makes the module unusable when the resources are indirectly created by other resources. e.g. a pod which is created by a deployment (or a pod created by a operator)

(no error text output, but this is the "register: result" json object:

```paste below
{
    "changed": false,
    "failed": false,
    "resources": []
}
```

gravesm:
##### SUMMARY

The wait logic in the k8s_info module immediately returns when no
resources are found, regardless whether a wait_timeout has been
specified. This expands the logic to wait when a name has been provided.
The case this is specifically meant to address is when querying for a
resource that is indirectly created by another resource, for example, a
pod created by a deployment or an operator.

The existing logic in every other case should remain as it was before.
This means if a query might return more than one resource--all pods with
some label, for example--the module will return immediately if no pods
are found, even if a wait_timeout has been provided.

Closes #344 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

k8s_info, k8s

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/module_utils/common.py
+++ b/plugins/module_utils/common.py
@@ -319,31 +319,58 @@ class K8sAnsibleMixin(object):
         if not field_selectors:
             field_selectors = []
 
+        result = None
         try:
-            result = resource.get(name=name,
-                                  namespace=namespace,
+            result = resource.get(name=name, namespace=namespace,
                                   label_selector=','.join(label_selectors),
                                   field_selector=','.join(field_selectors))
-            if wait:
-                satisfied_by = []
-                if isinstance(result, ResourceInstance):
-                    # We have a list of ResourceInstance
-                    resource_list = result.get('items', [])
-                    if not resource_list:
-                        resource_list = [result]
-
-                    for resource_instance in resource_list:
-                        success, res, duration = self.wait(resource, resource_instance,
-                                                           sleep=wait_sleep, timeout=wait_timeout,
-                                                           state=state, condition=condition)
-                        if not success:
-                            self.fail(msg="Failed to gather information about %s(s) even"
-                                          " after waiting for %s seconds" % (res.get('kind'), duration))
-                        satisfied_by.append(res)
-                    return dict(resources=satisfied_by, api_found=True)
-            result = result.to_dict()
-        except (openshift.dynamic.exceptions.BadRequestError, openshift.dynamic.exceptions.NotFoundError):
+        except openshift.dynamic.exceptions.BadRequestError:
             return dict(resources=[], api_found=True)
+        except openshift.dynamic.exceptions.NotFoundError:
+            if not wait or name is None:
+                return dict(resources=[], api_found=True)
+
+        if not wait:
+            result = result.to_dict()
+            if 'items' in result:
+                return dict(resources=result['items'], api_found=True)
+            return dict(resources=[result], api_found=True)
+
+        start = datetime.now()
+
+        def _elapsed():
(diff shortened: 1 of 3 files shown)


tima (approved):
LGTM but I'd defer to those who know K8s internals better than me.

plugins/module_utils/common.py
```diff
@@ -319,31 +319,58 @@ def kubernetes_facts(self, kind, api_version, name=None, namespace=None, label_s
...
+                except NotFoundError:
+                    pass
+                time.sleep(wait_sleep)
+            if result is None:
+                return dict(resources=[], api_found=True)
+
+        if isinstance(result, ResourceInstance):
+            satisfied_by = []
+            # We have a list of ResourceInstance
+            resource_list = result.get('items', [])
```
fabianvf, line 360: I wonder if it would be better to move the list handling behavior into the `wait` function instead of adding the `wait` behavior here

gravesm: Yeah, that's a good suggestion.

gravesm: So I started doing this, but I feel like I'm pulling on a stray thread in a sweater. Our current waiter implementation is really difficult to work with. As you point out, the logic for this is spread out over the code. The `wait` function doesn't support waiting on multiple resources. I can tack that on, but then the return values don't make sense.

I'd really like to refactor this part of the code. In addition to the problems listed above, there are some inconsistencies in how we're handling the timeout and a lack of clear intent for certain scenarios. My preference would be to do that work in a separate PR, partly so we don't block getting this bug fix into 1.2, but also so we can more easily track any regressions that might happen.

fabianvf: makes sense to me

jillr (approved):
This looks like a purely python logic change to me, LGTM from that perspective

PR #362, opened 2021-02-04
make has_plugin detection method more reliable

joschi36:
##### SUMMARY

This PR makes the plugin detection method more reliable.
The issue before was that when you had multiple plugins it wouldn't split at the right point. This method just loops through the list and searches the start of the string for the plugin name.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

helm

##### ADDITIONAL INFORMATION

Why the method before was a problem: With multiple plugins with different plugin name lengths installed the detection method failed.
```
NAME    	VERSION       	DESCRIPTION                                                                         
diff    	3.1.3         	Preview helm upgrade changes as a diff                                              
helm-git	0.8.1         	Get non-packaged Charts directly from Git.                                          
unittest	0.1.6-rancher1	Unit test for helm chart in YAML with ease to keep your chart functional and robust.
```
So my line is `diff\s\s\s\s\s\t3.1.3[...]` and the result variable `name` with the behavior before was `diff\s\s\s\s\s` which unfortunately wasn't equal to ´diff´
I hope I could explain it well, If you have an additional question please ask.

/cc @gravesm


gravesm:
@joschi36 thank you for the PR. I think the better solution here would be to just change it to `line.split(None, 1)`. If you searched for a plugin named `foo` but had one named `foobar` installed then using `startswith("foo")` would lead to a false positive.

Also, could you add a changelog fragment? https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment

joschi36:
Thanks for the feedback. I first also thought on doing that, however I choose to use the more simple variant. But yes you are totally right. I will update the PR with this change and also with the change log fragment.

joschi36:
@gravesm I have updated the PR

changelogs/fragments/362-helm-has_plugin-fix.yaml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - helm - make helm-diff plugin detection more reliable by splitting by any whitespace instead of explicit whitespace (`\s`)
```
gravesm, line 2: Just a couple minor nits. Since this gets turned into rst those should be double backticks. Could you also add the link to this PR at the end in parentheses? There are a few examples here: https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment.

Thanks for your work on this!

joschi36:
@gravesm I have fixed the double backticks.

gravesm (approved):
Thank you for your contribution, @joschi36!

PR #363, opened 2021-02-04
helm: use the diff generated before

joschi36:
##### SUMMARY
In the previous PR #355 @Akasurde has implemented optional helm check by the helm plugin diff.
I was unsure why we would not use this for --diff in Ansible. This would be very useful.
I have added a gist with a simple module I build which is actually using it.
Also, I think this PR needs some more love, but first I wanted to gather a little bit of feedback from the original creator @Akasurde and @LucasBoisserie 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

Helm

##### ADDITIONAL INFORMATION
Also, I sorted parameters in `exit_json`

### DIFF
--- a/plugins/modules/helm.py
+++ b/plugins/modules/helm.py
@@ -440,7 +440,7 @@ def helmdiff_check(module, helm_cmd, release_name, chart_ref, release_values,
             cmd += " -f=" + values_file
 
     rc, out, err = run_helm(module, cmd)
-    return len(out.strip()) > 0
+    return len(out.strip()) > 0, out
 
 
 def default_check(release_status, chart_info, values=None, values_files=None):
@@ -513,6 +513,7 @@ def main():
         module.fail_json(msg=missing_required_lib("yaml"), exception=IMP_YAML_ERR)
 
     changed = False
+    diff = ""
 
     bin_path = module.params.get('binary_path')
     chart_ref = module.params.get('chart_ref')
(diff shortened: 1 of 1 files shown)


tima:
Thanks for the PR @joschi36. CI failures aside I had a couple of comments.

Should the presence of the diff be conditional on the presence of `--diff`? I'm not seeing that here. Also, since `helm-diff` is optional to this module, what is returned if `--diff` is used?

joschi36:
Thanks @tima for the quick feedback.

First sorry I've forgotten to add the gist I was talking about: https://gist.github.com/joschi36/f7df3a0aac460cebe11c36ec1b9a03bf
That's the module we are currently using in production.

> Should the presence of the diff be conditional on the presence of --diff?

I'm not sure if this is needed. If the `--diff` parameter is not present the diff is not showed. I'm not sure if it should be conditional.

> Also, since helm-diff is optional to this module, what is returned if --diff is used?

No diff as before. When variable diff is empty it will result in:
`diff={'prepared': ''},` in `exit_json` and just no diff is shown.

I hope I understood your questions right, it's getting late here :smile:

plugins/modules/helm.py
```diff
@@ -573,7 +574,7 @@ def main():
         else:
 
             if has_plugin(helm_cmd_common, "diff") and not chart_repo_url:
-                would_change = helmdiff_check(module, helm_cmd_common, release_name, chart_ref,
+                would_change, diff = helmdiff_check(module, helm_cmd_common, release_name, chart_ref,
                                               release_values, values_files, chart_version, replace)
```
gravesm, line 578: The indentation here needs to be corrected in order to address the linting error.

gravesm (changes requested):
Thanks for the PR @joschi36. Aside from the linting error we'll need a [changelog fragment](https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment).

This looks good, though, and is a nice addition!

Akasurde:
@joschi36 Thanks for the contribution. Could you open a new PR against https://github.com/ansible-collections/kubernetes.core? Thanks.

PR #368, opened 2021-02-11
Add support for helm template

### LINKED ISSUE #367: A module for rendering Helm3 templates
##### SUMMARY

Create a module for rendering Helm3 templates --  the functional equivalent of the [`helm template`](https://helm.sh/docs/helm/helm_template/) command.

This comes out of discussion in #285.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

`helm_template`

##### ADDITIONAL INFORMATION

Sometimes a chart needs to be different under under different conditions so a template is used. While an Ansible content developer could use the `template` module an Jinja2, some may already have or be forced to work with Helm charts. 
 Ansible should address this scenario in a more Helm-native way with a module that is the functional equivalent of the `helm template command.

This module should use parameters like `helm` and other modules for cohesive and consistent operation with Helm3 automation.

gravesm:
##### SUMMARY

This commit adds basic support for helm template. The functionality
allows one to either render a chart's templates to a specific directory,
or capture the rendered templates as a string of concatenated yaml
documents.

Closes #367 

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

helm_template

##### ADDITIONAL INFORMATION

Sorry, for the extensive diff on `tests_chart.yml` but I really wanted to fix our test teardown. Aside from indentation changes, the only lines in that file that are new are 3-11 and 325-349.

```paste below

```

### DIFF
--- a/molecule/default/roles/helm/tasks/tests_chart.yml
+++ b/molecule/default/roles/helm/tasks/tests_chart.yml
@@ -332,6 +332,12 @@
           - "{{ role_path }}/files/values.yaml"
       register: result
 
+    - assert:
+        that:
+          - result is not failed
+          - result.rc == 0
+          - result.command is match("{{ helm_binary }} template {{ chart_source }}")
+
     - name: Check templates created
       stat:
         path: "{{ temp_dir }}/{{ chart_test }}/templates"
--- a/molecule/default/roles/helm/tasks/tests_chart.yml
+++ b/molecule/default/roles/helm/tasks/tests_chart.yml
@@ -334,6 +334,7 @@
 
     - assert:
         that:
+          - result is changed
           - result is not failed
           - result.rc == 0
           - result.command is match("{{ helm_binary }} template {{ chart_source }}")
--- a/molecule/default/roles/helm/tasks/tests_chart.yml
+++ b/molecule/default/roles/helm/tasks/tests_chart.yml
@@ -1,321 +1,358 @@
 ---
-- name: Check helm_info empty
-  helm_info:
-    binary_path: "{{ helm_binary }}"
-    name: test
-    namespace: "{{ helm_namespace }}"
-  register: empty_info
+- name: Chart tests
+  block:
+    - name: Create temp directory
+      tempfile:
+        state: directory
+      register: tmpdir
 
-- name: "Assert that no charts are installed with helm_info"
-  assert:
-    that:
-      - empty_info.status is undefined
+    - name: Set temp directory fact
+      set_fact:
+        temp_dir: "{{ tmpdir.path }}"
 
-- name: "Install fail {{ chart_test }} from {{ source }}"
-  helm:
-    binary_path: "{{ helm_binary }}"
-    name: test
-    chart_ref: "{{ chart_source }}"
-    chart_version: "{{ chart_source_version | default(omit) }}"
-    namespace: "{{ helm_namespace }}"
-  ignore_errors: yes
-  register: install_fail
+    - name: Check helm_info empty
+      helm_info:
+        binary_path: "{{ helm_binary }}"
+        name: test
+        namespace: "{{ helm_namespace }}"
+      register: empty_info
 
-- name: "Assert that Install fail {{ chart_test }} from {{ source }}"
-  assert:
-    that:
-      - install_fail is failed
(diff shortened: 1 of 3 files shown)


plugins/modules/helm_template.py
```diff
@@ -0,0 +1,213 @@
...
+  - Render chart templates to an output directory or as text of concatenated yaml documents.
+
+options:
+  binary_path:
+    description:
+      - The path of a helm binary to use.
+    required: false
+    type: path
+  chart_ref:
+    description:
```
jillr, line 31: Would this option be one of the listed items?  It might be more obvious to folks who are familiar with Helm but IMO we should say something like, "Chart reference with a repo prefix. Can be one of the following:" and then list the next 3 options.

plugins/modules/helm_template.py
```diff
@@ -0,0 +1,213 @@
...
+      - Chart reference with repo prefix.
+      - Path to a packaged chart.
+      - Path to an unpacked chart directory.
+      - Absolute URL.
+    required: true
+    type: path
+  chart_repo_url:
+    description:
+      - Chart repository URL where to locate the requested chart.
```
jillr, line 40:
```suggestion
      - Chart repository URL where the requested chart is located.
```
Nit

plugins/modules/helm_template.py
```diff
@@ -0,0 +1,213 @@
...
+stderr:
+  type: str
+  description: Full C(helm) command stderr, in case you want to display it or examine the event log.
+  returned: always
+  sample: ''
+command:
+  type: str
+  description: Full C(helm) command run by this module, in case you want to re-run the command outside the module or debug a problem.
+  returned: always
+  sample: helm upgrade ...
```
jillr, line 115: Is `helm upgrade` a possible `command` for this module to return or should this be a `helm template` command?

jillr (commented):
Couple of small docs nits, the Python parts of the module lgtm

plugins/modules/helm_template.py
gravesm: "Chart reference with a repo prefix" is actually one of the options (there are four). I can add a "for example" though, which should help clarify.

tima (approved):
LGTM

molecule/default/roles/helm/tasks/tests_chart.yml
```diff
@@ -1,321 +1,358 @@
...
+
+    - name: Render templates
+      helm_template:
+        binary_path: "{{ helm_binary }}"
+        chart_ref: "{{ chart_source }}"
+        chart_version: "{{ chart_source_version | default(omit) }}"
+        output_dir: "{{ temp_dir }}"
+        values_files:
+          - "{{ role_path }}/files/values.yaml"
+      register: result
```
jillr, line 333: It would be nice to have some assertions for this task.  is not failed, result.rc value, and result.command should be straightforward, depending on the reproducability and complexity of result.stdout there should be something we can assert about the shape or contents that.

jillr (commented):
We should add something to the module docs about idempotency, we talked in slack and it sounds like the CLI is not idempotent so we should just note that so that users expecting Ansible's usual behaviour aren't surprised.

Could we add check mode support to this module?  I'm not really familiar enough with how folks use Helm and would expect that to work, but in theory we could still build the command with `--dry-run` (or even just store the rendered yaml to stdout?).

goneri (approved):

molecule/default/roles/helm/tasks/tests_chart.yml
gravesm: Unfortunately, examining stdout here will be difficult. This test suite gets run multiple times with different charts so the output will be different for each one. The same thing goes for examining `result.command` to some extent. I've added a check for as much of the command as we can reliably predict.

plugins/modules/helm_template.py
```diff
@@ -0,0 +1,220 @@
...
+
+    if not check_mode:
+        rc, out, err = run_helm(module, tmpl_cmd)
+    else:
+        out = err = ""
+        rc = 0
+
+    module.exit_json(
+        failed=False,
+        changed=False,
```
jillr, line 211: Sorry I missed this the first time around, shouldn't changed always be True?

jillr (approved):
LGTM - Thanks @gravesm

PR #392, opened 2021-03-17
enable the AnsibleTurboModule

goneri:
Use AnsibleTurboModule when the `cloud.common` collection is available.


plugins/module_utils/ansiblemodule_turbo.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+try:
+    from ansible_collections.cloud.common.plugins.module_utils.turbo.module import (
+        AnsibleTurboModule as AnsibleModule,
+    )  # noqa: F401
+    AnsibleModule.collection_name = "community.kubernetes"
+except (ImportError, ModuleNotFoundError):
+    from ansible.module_utils.basic import AnsibleModule  # noqa: F401
+from ansible.module_utils.basic import AnsibleModule  # noqa: F401
```
Akasurde, line 13: isn't `AnsibleModule` redundant?

gravesm:
The CI is now failing for three reasons:
1. 2.7 incompatibility. Maybe we can just drop 2.7 from the test matrix since it sounds like we may not be supporting it?
2. A legitimate linting error in cloud.common. If there's a way to ignore errors in cloud.common I'd be fine with that as it's not really this collection's problem. It should probably be fixed in cloud.common, though.
3. The known issue with molecule 3.3.0. I'll pin our molecule dependency in a separate PR and we can rebase to fix this problem.

gravesm:
Unless I'm missing something `ansiblemodule.py` and `ansiblemodule_turbo.py` are exactly the same. Is there a reason we need both?

goneri:
See: https://github.com/ansible-collections/kubernetes.core/pull/68

PR #395, opened 2021-03-23
k8s: fix get_api_client usage in inventory plugin

Akasurde:
##### SUMMARY

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
changelogs/fragments/k8s_inventory.yml
plugins/inventory/k8s.py

### DIFF
--- a/plugins/module_utils/common.py
+++ b/plugins/module_utils/common.py
@@ -118,7 +118,7 @@ def configuration_digest(configuration):
     return digest
 
 
-def get_api_client(module=None):
+def get_api_client(module=None, **kwargs):
     auth = {}
 
     def _raise_or_fail(exc, msg):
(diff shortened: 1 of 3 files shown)


plugins/module_utils/common.py
```diff
@@ -118,7 +118,7 @@ def configuration_digest(configuration):
     return digest
 
 
-def get_api_client(module=None):
+def get_api_client(module=None, **kwargs):
```
goneri, line 121: Hi @Akasurde, can you explain why you need this? These parameters will be ignore.

goneri (changes requested):

plugins/module_utils/common.py
Akasurde: Inventory plugin calls `get_api_client` with args, which fails the inventory plugin.

goneri: Yes, but these arguments will be ignored anyway. Maybe we should just remove them?

Akasurde: No they are used.

goneri (approved):
I would prefer to see the molecule fix in a different PR.

gravesm (approved):

PR #400, opened 2021-03-29
helm: handle multiline output

Akasurde:
##### SUMMARY

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/helm.py

### DIFF
--- a/plugins/modules/helm_plugin.py
+++ b/plugins/modules/helm_plugin.py
@@ -198,6 +198,8 @@ def main():
                     continue
                 name, dummy, dummy = line.split('\t', 3)
                 name = name.strip()
+                if name == "":
+                    continue
                 if name == plugin_name:
                     found = True
                     break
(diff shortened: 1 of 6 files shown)


plugins/modules/helm_plugin.py
```diff
@@ -198,6 +198,8 @@ def main():
                     continue
                 name, dummy, dummy = line.split('\t', 3)
                 name = name.strip()
+                if name == "":
```
abikouo, line 201: what if we put this piece of code which is common to the 3 modules into a function in module_utils/helm.py ?

Akasurde: Yes, we can. Thanks,

gravesm:
@Akasurde could you re-open this in kubernetes.core? Also, would be good to address @abikouo's comment, then we can get this merged.

Akasurde:
superseded by https://github.com/ansible-collections/kubernetes.core/pull/64

PR #417, opened 2021-04-14
k8s: handle ``fail_json`` API from inventory plugin

Akasurde:
##### SUMMARY

* inventory plugin does not have module object so fail_json
  will not work. Use AnsibleError instead of ``fail_json``

Fixes: #57

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
changelogs/fragments/fail_json.yml


gravesm:
I think this should be filed in kubernetes.core.

Akasurde:
superseded by https://github.com/ansible-collections/kubernetes.core/pull/59

PR #442, opened 2021-08-19
Update README

Akasurde:
* Update README for redirected link
* Enable Repo Lockdown in GitHub workflow

### DIFF
--- a/.github/workflows/lockdown.yml
+++ b/.github/workflows/lockdown.yml
@@ -1,0 +1,27 @@
+name: 'Repo Lockdown'
+
+on:
+  issues:
+    types: opened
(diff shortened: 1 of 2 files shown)


abikouo (approved):

gravesm (commented):
Thanks, this is a good idea. Do you think we should add an exclusion for PRs with a specific label, like "keep" or something? This would make it easier for us to still make changes if we have to.

Akasurde:
> s is a good idea. Do you think we should add an exclusion for PR

Yes definitely. Let me do the changes.

gravesm (approved):

